PR #42, opened 2023-05-29
Handle deeplinks from Staging

ipavlidakis:
### 🎯 Goal

Provide parity on deeplink handling between DemoApp and Dogfood

### Manual Testing

Simulator:
- Run the app
- From terminal execute ```xcrun simctl openurl booted "streamvideo://video?id=<call-id>"```

Device:
- Install the app
- On Safari Paste & Go to ```"streamvideo://video?id=<call-id>"```

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/CallView.swift
+++ b/DemoApp/CallView.swift
@@ -10,8 +10,8 @@ import Intents
 struct CallView: View {
 
     private var callId: String
+    @Injected(\.streamVideo) var streamVideo
     @StateObject var viewModel: CallViewModel
-    
     @ObservedObject var appState = AppState.shared
     
     init(callId: String) {
@@ -48,8 +48,12 @@ struct CallView: View {
         }
 
         Task {
-            try await appState.streamVideo?.connect()
-            viewModel.joinCall(callId: callId, type: callType)
+            await MainActor.run {
+                Task {
+                    try await streamVideo.connect()
+                    viewModel.joinCall(callId: callId, type: callType)
+                }
+            }
         }
     }
 }
--- a/DemoApp/CallView.swift
+++ b/DemoApp/CallView.swift
@@ -33,16 +33,16 @@ struct CallView: View {
                     }
                 }
             )
-            .onAppear { joinCall(with: callId) }
+            .onAppear { joinCallIfNeeded(with: callId) }
             .onReceive(appState.$deeplinkInfo) { deeplinkInfo in
                 if deeplinkInfo != .empty {
-                    joinCall(with: deeplinkInfo.callId, callType: deeplinkInfo.callType)
+                    joinCallIfNeeded(with: deeplinkInfo.callId, callType: deeplinkInfo.callType)
                     appState.deeplinkInfo = .empty
                 }
             }
     }
 
-    private func joinCall(with callId: String, callType: String = .default) {
+    private func joinCallIfNeeded(with callId: String, callType: String = .default) {
         guard !callId.isEmpty, viewModel.callingState == .idle else {
             return
         }
--- a/DemoApp/CallView.swift
+++ b/DemoApp/CallView.swift
@@ -8,16 +8,15 @@ import StreamVideoSwiftUI
 import Intents
 
 struct CallView: View {
-    
+
+    private var callId: String
     @StateObject var viewModel: CallViewModel
     
     @ObservedObject var appState = AppState.shared
     
     init(callId: String) {
         _viewModel = StateObject(wrappedValue: CallViewModel())
-        if !callId.isEmpty, viewModel.callingState == .idle {
-            viewModel.joinCall(callId: callId, type: .default)
-        }
+        self.callId = callId
     }
         
     var body: some View {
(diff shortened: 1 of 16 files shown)


DemoApp/DeeplinkAdapter.swift
```diff
@@ -0,0 +1,42 @@
...
+}
+
+struct DeeplinkAdapter {
+    var baseURL: URL
+
+    func canHandle(url: URL) -> Bool {
+        url.absoluteString.contains(baseURL.absoluteString)
+    }
+
+    func handle(url: URL, completion: (DeeplinkInfo, User) -> Void) {
```
martinmitrevski, line 25: Why do we need completion here? Can't we just directly return the deeplink info?

DemoApp/HomeView.swift
```diff
@@ -120,6 +120,11 @@ struct HomeView: View {
         .onReceive(appState.$activeCall) { call in
             viewModel.setActiveCall(call)
         }
+        .observeAndHandleDeeplinks(
```
martinmitrevski, line 123: Can we add this in the parent view, instead of in two places?

DemoApp/DemoAppUtils.swift
```diff
@@ -14,3 +17,25 @@ extension URL {
...
+    func observeAndHandleDeeplinks(
+        _ viewModel: CallViewModel,
+        deeplinkInfoPublisher: AnyPublisher<DeeplinkInfo, Never>,
+        resetAppState: @escaping () -> Void
+    ) -> some View {
+        self.onReceive(deeplinkInfoPublisher) { deeplinkInfo in
+            if deeplinkInfo != .empty {
+                /// https://github.com/apple/swift/pull/60688
+                Task { @MainActor in
+                    viewModel.joinCall(
```
martinmitrevski, line 32: I think it would be more complicated than this:
- What happens if we are not logged in? 
- What happens if we are still not connected to the web socket?
- What happens if the token is expired?

We will also probably support a case where the token is provided in the URL, or we join anonymously or as guest.

martinmitrevski (commented):
Before merging, we should also make it work with the website team.

DemoApp/DemoAppUtils.swift
ipavlidakis: > What happens if we are not logged in?
I see that `HomeView` is only available to `loggedIn` users. In the case you are not `loggedIn` the view won't be on the hierarchy (or even created) so there shouldn't be any issue there.

> What happens if we are still not connected to the web socket?
> What happens if the token is expired?

Don't we already handle those scenarios in the `joinCall` as prerequisites? 

> We will also probably support a case where the token is provided in the URL, or we join anonymously or as guest.
I'm not sure how (probably i don't have enough context) this is going to affect it

DemoApp/CallView.swift
```diff
@@ -8,16 +8,15 @@ import StreamVideoSwiftUI
...
 
 struct CallView: View {
-    
+
+    private var callId: String
     @StateObject var viewModel: CallViewModel
     
     @ObservedObject var appState = AppState.shared
     
-    init(callId: String? = nil) {
```
martinmitrevski, line 16: Why is this not optional anymore? It should be used only for deeplinking.

DemoApp/CallView.swift
```diff
@@ -34,6 +33,24 @@ struct CallView: View {
                     }
                 }
             )
+            .onAppear { joinCall(with: callId) }
```
martinmitrevski, line 36: This can be confusing for our integrators, since it looks like we always join a call on appear. Maybe we can update the method name to `joinCallIfNeeded`?

DemoApp/StreamVideoSwiftUIApp.swift
```diff
@@ -78,30 +86,38 @@ struct StreamVideoSwiftUIApp: App {
                 }
             }
         )
+        setUp(streamVideo: streamVideo, deeplinkInfo: deeplinkInfo, user: credentials.userInfo)
+    }
+
+    private func handleGuestUser(deeplinkInfo: DeeplinkInfo) async throws {
+        let user = {
+            guard let currentUser = appState.currentUser, currentUser.id == currentUser.name else {
+                return User.guest(UUID().uuidString)
```
martinmitrevski, line 95: as discussed, let's shorten this, it will not look good in the UI.

martinmitrevski (commented):
Looks good 👍  Added some small comments. Also, I still need to test it.

DemoApp/CallView.swift
ipavlidakis: I preferred emptyness than nullability. This is aligned with the DeeplinkInfo that isn't nil but instead `.empty`. I consider it more "type-safe" and explicit but i'm happy to hear your thoughts if you prefer nullability.

martinmitrevski: Ok, let's keep it like this then.

martinmitrevski (approved):
Works great ✅

PR #53, opened 2023-06-15
Sync xcodeproj references as part of code generation

testableapple:
Resolves https://github.com/GetStream/ios-issues-tracking/issues/476


fastlane/Fastfile
```diff
@@ -288,6 +288,49 @@ private_lane :emerge_upload do |options|
   )
 end
 
+lane :sync_xcodeproj_references do
+  Dir.chdir('..') do
+    project = Xcodeproj::Project.open(xcode_project)
+
+    untracked_files = sh('git status -s', log: false).split("\n").map(&:strip)
+    deleted_files = untracked_files.select { |f| f.start_with?('D ') && f.end_with?('.swift') }.map { |f| f.split.drop(1).join }
+    new_files = untracked_files.reject { |f| f.start_with?('M ', 'D ') }.map { |f| f.split.drop(1).join }
```
tbarbugli, line 297: selecting known prefixes is safer (?? and A)

fastlane/Fastfile
```diff
@@ -288,6 +288,49 @@ private_lane :emerge_upload do |options|
   )
 end
 
+lane :sync_xcodeproj_references do
```
tbarbugli, line 291: probably good to add a couple params here so we can use it for protobuf and for openapi

fastlane/Fastfile
testableapple: Updated the PR, check it out

fastlane/Fastfile
testableapple: Hmm, but why do we need it? The current script deals
- with all new `.swift` files which are not referenced in `.xcodeproj`
- and with all deleted `.swift` files

PR #75, opened 2023-06-27
Fix View Change button not clickable when participant image contains …

ipavlidakis:
### 🎯 Goal

- View button should be clickable when non primary participant thumbnail's location is topTrailing and the thumbnail contains an image.
- Minimize(back) button should be clickable when non primary participant thumbnail's location is topLeading and the thumbnail contains an image.

### 🛠 Implementation

Calling `.aspectRatio(contentMode: .fill)` was causing the view to extend outside its bounds and overlap with topBar's button areas. Additionally, setting the contentMode of a view doesn't produce the same effect as setting the contentMode of an Image.

The right way to do it by calling the following:
```swift
Image(url: ...)
    .resizable()
    .aspectRation(contentMode: .fill)
```

In our case, it seems like this wasn't event necessary as we are using LazyImage instead of Image and we are already setting the resizingMode to aspectFill.

### 🎨 Showcase

https://github.com/GetStream/stream-video-swift/assets/472467/0bd2ea53-20f6-40ed-a79b-a96ff93585bf

https://github.com/GetStream/stream-video-swift/assets/472467/3b8d3bcf-57a6-478d-b167-d7c9d50f3824

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

- [x] Comparison screenshots added for visual changes

### DIFF
--- a/Sources/StreamVideoSwiftUI/CallView/CallView.swift
+++ b/Sources/StreamVideoSwiftUI/CallView/CallView.swift
@@ -59,26 +59,24 @@ public struct CallView<Factory: ViewFactory>: View {
                     }
                 }
                                 
-                TopRightView {
-                    VStack(alignment: .trailing, spacing: padding) {
-                        viewFactory.makeCallTopView(viewModel: viewModel)
-                        
-                        if viewModel.screensharingSession == nil, viewModel.participantsLayout == .grid {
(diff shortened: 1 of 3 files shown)


martinmitrevski (approved):
nice catch 👍

PR #85, opened 2023-06-29
Implement retry mechanism

martinmitrevski:


Sources/StreamVideo/WebRTC/PeerConnection.swift
```diff
@@ -27,8 +27,9 @@ class PeerConnection: NSObject, RTCPeerConnectionDelegate, @unchecked Sendable {
     private var pendingIceCandidates = [RTCIceCandidate]()
     private var publishedTracks = [TrackType]()
     private var screensharingStreams = [RTCMediaStream]()
+    private let badConnectionStates: [RTCIceConnectionState] = [.disconnected, .failed, .closed]
```
ipavlidakis, line 30: Can we make this one a set to take advantage of faster lookup operations?

Sources/StreamVideo/WebRTC/Retries.swift
```diff
@@ -0,0 +1,77 @@
...
+    
+    static func neverGonnaGiveYouUp(_ condition: @escaping () -> Bool) -> RetryPolicy {
+        RetryPolicy(
+            maxRetries: 30,
+            delay: { delay(retries: $0) },
+            runPrecondition: condition
+        )
+    }
+    
+    static func quickDelay(retries: Int) -> TimeInterval {
```
ipavlidakis, line 70: Why are we passing retries in?

Sources/StreamVideo/WebRTC/Retries.swift
```diff
@@ -0,0 +1,77 @@
...
+}
+
+struct RetryPolicy {
+    let maxRetries: Int
+    let delay: (Int) -> TimeInterval
+    var runPrecondition: () -> Bool = { true }
+}
+
+extension RetryPolicy {
+    static let fastAndSimple = RetryPolicy(maxRetries: 3) { quickDelay(retries: $0) }
```
ipavlidakis, line 52: I think we should make this one a no-delay operation as it's being described in the doc
```suggestion
    static let fastAndSimple = RetryPolicy(maxRetries: 3) { 0 }
```

Sources/StreamVideo/WebRTC/WebRTCClient.swift
```diff
@@ -431,25 +443,33 @@ class WebRTCClient: NSObject {
         await assignTracksToParticipants()
         let state = await self.state.connectionState
         if state == .connected {
-            await updateParticipantsSubscriptions()
+            try? await updateParticipantsSubscriptions()
```
ipavlidakis, line 446: Why not `try`?

StreamVideoTests/WebRTC/Retries_Tests.swift
```diff
@@ -0,0 +1,119 @@
...
+@testable import StreamVideo
+import WebRTC
+import XCTest
+
+final class Retries_Tests: XCTestCase {
+    
+    let dummyRequest = URLRequest(url: URL(string: "https://test.com")!)
+    let dummyData = Data()
+    let dummyError = ClientError.NetworkError()
+    var dummyState = "dummy"
```
ipavlidakis, line 14:
```suggestion
    private var dummyRequest: URLRequest! = URLRequest(url: URL(string: "https://test.com")!)
    private var dummyData: Data! = Data()
    private var dummyError: Error! = ClientError.NetworkError()
    private var dummyState: String! = "dummy"
    
    func tearDown() throws async {
        dummyRequest = nil
        dummyData = nil
        dummyError = nil
        dummyState = nil
        try await super.tearDown()
    }
```

or make them private static

Sources/StreamVideo/WebRTC/PeerConnection.swift
martinmitrevski: it's only 3 elements, doesn't make any difference

Sources/StreamVideo/WebRTC/Retries.swift
martinmitrevski: in case we want to do exponential backoff in the future

Sources/StreamVideo/WebRTC/WebRTCClient.swift
martinmitrevski: because we still want to update the local state (the next lines) even if the request failed

PR #92, opened 2023-07-03
Improved thread safe usage of the CallState in SwiftUI

martinmitrevski:


Sources/StreamVideo/Call.swift
```diff
@@ -546,13 +548,17 @@ public class Call: @unchecked Sendable, WSEventsSubscriber {
     //MARK: - Internal
     
     internal func update(reconnectionStatus: ReconnectionStatus) {
-        if reconnectionStatus != self.state.reconnectionStatus {
-            self.state.reconnectionStatus = reconnectionStatus
+        executeOnMain {
+            if reconnectionStatus != self.state.reconnectionStatus {
```
ipavlidakis, line 552: Why strongly capturing self in all instances of `executeOnMain`?

ipavlidakis (approved):

PR #97, opened 2023-07-03
Bump Xcode version on CI

testableapple:


martinmitrevski (approved):
Let's make sure the snapshot tests are passing before merging.

PR #154, opened 2023-09-07
Prevent converting buffers when PiP is inactive

martinmitrevski:


Sources/StreamVideoSwiftUI/Utils/PiPHandler.swift
ipavlidakis, line 1: Shall we nullify the `feedFrames` when we stop PiP

martinmitrevski: that's already done in the `cleanUp` method (called when we stop pip)

ipavlidakis: Oh my bad. I missed it even though i looked :P

ipavlidakis (approved):

PR #156, opened 2023-09-12
Implement stats reports

martinmitrevski:
### 🔗 Issue Links

_Provide all JIRA tickets and/or GitHub issues related to this PR, if applicable._

### 🎯 Goal

Expose call stats via the call state.

This PR is only LLC change, the updates in the demo app would be in a separate PR.

### ☑️ Contributor Checklist

### 🎁 Meme

Alt Text


Sources/StreamVideo/Controllers/CallController.swift
ipavlidakis, line 1: May i suggest a functional approach on this one using Combine (which i find to be more readable)

```swift
import Combine

...

private var statsCancellable: AnyCancellable?

...

func cleanUp() {
...
statsCancellable?.cancel()
statsCancellable = nil
...
}

...

private func setupStatsTimer() {
    statsCancellable?.cancel()
    statsCancellable = Foundation.Timer.publish(
        every: 5,
        on: .main,
        in: .default
    ).sink { [weak self] _ in self?.collectStats() }
}
```

Sources/StreamVideo/WebRTC/StatsReporter.swift
```diff
@@ -0,0 +1,168 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import WebRTC
+
+enum StatsReporter {
```
ipavlidakis, line 8: Why not a have all those calculations as part of the CallStatsReport's init `.init(reports: [RTCStatisticsReport?], datacenter: String)` and all the calculations happening in the init? (reduce the static load)

ipavlidakis (commented):
How bound are we on the `Stats` String? Short strings reduce how explicit something is and increase the chance for confusion imo. How do you feel about using the `statistics` rather than stats?

Sources/StreamVideo/Controllers/CallController.swift
martinmitrevski: Yep, good idea. Done, I had to add also:
```
.autoconnect()
.receive(on: RunLoop.main)
```

Sources/StreamVideo/WebRTC/StatsReporter.swift
martinmitrevski: `RTCStatisticsReport` can't be instantiated, which makes it impossible to test. I have it split so I can test some chunks easily (the parts where we work with dictionaries).

DemoApp/Sources/Components/Stats/ParticipantStatsView.swift
```diff
@@ -0,0 +1,99 @@
...
+                    .padding(.all, spacing)
+                    .background(Color(colors.background1))
+                    .cornerRadius(8)
+            }
+
+        }
+        .frame(width: 2 * itemSize + 4 * spacing)
+        .padding()
+        .background(colors.callControlsBackground)
+        .cornerRadius(8)
```
ipavlidakis, line 62:
```suggestion
.clipped()
.clipShape(RoundedRectangle(cornerRadius: 8))
```

I think it's better to use the clipShape to avoid artifacts getting rendered outside of the bound box

DemoApp/Sources/Components/Stats/ParticipantStatsView.swift
```diff
@@ -0,0 +1,99 @@
...
+    
+    private var itemSize: CGFloat {
+        if compacted {
+            return 72
+        } else {
+            return 100
+        }
+    }
+    
+    private var valueFont: Font {
```
ipavlidakis, line 80: Shall we use our fonts here?

DemoApp/Sources/Components/Stats/ParticipantStatsViewModel.swift
```diff
@@ -0,0 +1,102 @@
...
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Combine
+import StreamVideo
+import StreamVideoSwiftUI
+import SwiftUI
+
+@MainActor
+class ParticipantStatsViewModel: ObservableObject {
```
ipavlidakis, line 11: nit: `final`

DemoApp/Sources/Components/Stats/ParticipantStatsViewModel.swift
```diff
@@ -0,0 +1,102 @@
...
+        call: Call,
+        participant: CallParticipant
+    ) {
+        self.call = call
+        self.participant = participant
+        self.subscribeToStatsUpdates()
+    }
+
+    private func subscribeToStatsUpdates() {
+        call.state.$statsReport.sink { [weak self] statsReport in
```
ipavlidakis, line 41: Can we move this one in background with `receive(on: DispatchQueue.global())` as the calculation isn't really required to be on MainThread (UI changes can be moved to Main when necessary)

DemoApp/Sources/Components/Stats/RawStatsView.swift
```diff
@@ -0,0 +1,75 @@
...
+        
+    var statsReport: CallStatsReport?
+    
+    var body: some View {
+        NavigationView {
+            if statsReport != nil {
+                ScrollView {
+                    LazyVStack(alignment: .leading) {
+                        Text("Subscriber stats")
+                            .font(.headline)
```
ipavlidakis, line 19: nit: Can we use our fonts?

Sources/StreamVideo/Controllers/CallController.swift
```diff
@@ -454,6 +460,26 @@ class CallController {
...
     
+    private func setupStatsTimer() {
+        statsCancellable?.cancel()
+        statsCancellable = Foundation.Timer.publish(
+            every: 5,
+            on: .main,
+            in: .default
+        )
+        .autoconnect()
+        .receive(on: RunLoop.main)
```
ipavlidakis, line 471: Probably better to use `DispatchQueue.main` to ensure responsiveness ([reference](https://www.avanderlee.com/combine/runloop-main-vs-dispatchqueue-main/))

martinmitrevski: good point 👍

DemoApp/Sources/Components/Stats/ParticipantStatsViewModel.swift
martinmitrevski: hmm, there's no calculation here, just setting the values. The calculation is already done on a background thread in the LLC.

DemoApp/Sources/Components/Stats/ParticipantStatsView.swift
martinmitrevski: done

DemoApp/Sources/Components/Stats/ParticipantStatsView.swift
martinmitrevski: done

ipavlidakis (approved):

PR #163, opened 2023-09-26
[Fix]Deeplink handling

ipavlidakis:
### Issue
Resolves https://github.com/GetStream/ios-issues-tracking/issues/571

### 🎯 Goal

Ensure deeplinks are working consistently

### 📝 Summary

A few issues discovered:
- Association file was broken
- Missing the implementation method that iOS seems to call for universal links but not for custom schemes

### 🛠 Implementation

Implemented UITests that validate deeplink handling for 
- Universal links (via the smart banner open button)
- Custom URL schemes

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/Resources/Entitlements/DemoApp-Debug.entitlements
+++ b/DemoApp/Resources/Entitlements/DemoApp-Debug.entitlements
@@ -6,8 +6,7 @@
 	<string>development</string>
 	<key>com.apple.developer.associated-domains</key>
 	<array>
-		<string>applinks:staging.getstream.io</string>
-		<string>applinks:getstream.io</string>
+        <string>applinks:getstream.io</string>
 	</array>
 	<key>com.apple.security.application-groups</key>
 	<array>
--- a/DemoApp/Sources/Components/AppEnvironment.swift
+++ b/DemoApp/Sources/Components/AppEnvironment.swift
@@ -229,7 +229,7 @@ extension AppEnvironment {
         case .debug:
             return [.staging, .pronto, .production]
         case .test:
-            return [.staging]
+            return [.staging, .pronto, .production]
         case .release:
             return [.production]
         }
--- a/DemoApp/Sources/Components/MemoryLogDestination/OSLogDestination.swift
+++ b/DemoApp/Sources/Components/MemoryLogDestination/OSLogDestination.swift
@@ -1,0 +1,14 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamVideo
+import OSLog
+
+final class OSLogDestination: BaseLogDestination {
+
+    override func write(message: String) {
+        os_log("%@", message)
+    }
+}
--- a/DemoApp/Sources/Components/MemoryLogDestination/OSLogDestination.swift
+++ b/DemoApp/Sources/Components/MemoryLogDestination/OSLogDestination.swift
@@ -9,6 +9,6 @@ import OSLog
 final class OSLogDestination: BaseLogDestination {
 
     override func write(message: String) {
-        os_log("%@", message)
+        os_log("%{public}s", message)
     }
 }
--- a/DemoApp/Sources/Views/CallView/CallingView/DetailedCallingView.swift
+++ b/DemoApp/Sources/Views/CallView/CallingView/DetailedCallingView.swift
@@ -107,7 +107,6 @@ struct DetailedCallingView: View {
             .alignedToReadableContentGuide()
         }
         .navigationViewStyle(StackNavigationViewStyle())
-        //        .navigationBarHidden(true)
         .onAppear() {
             CallService.shared.registerForIncomingCalls()
         }
--- a/DemoApp/Sources/Views/CallView/CallingView/DetailedCallingView.swift
+++ b/DemoApp/Sources/Views/CallView/CallingView/DetailedCallingView.swift
@@ -10,15 +10,15 @@ import StreamVideoSwiftUI
 
 struct DetailedCallingView: View {
     @ObservedObject var viewModel: CallViewModel
-    
+
     @Injected(\.streamVideo) var streamVideo
-    
+
     private let imageSize: CGFloat = 32
-        
+
     @State private var callId = ""
-    
+
     @State private var callAction = CallAction.startCall
-    
+
     @State private var callFlow: CallFlow = .joinImmediately
 
     var participants: [User] {
@@ -28,13 +28,13 @@ struct DetailedCallingView: View {
         }
         return participants
     }
-    
+
     @State var selectedParticipants = [User]()
     @State var incomingCallInfo: IncomingCall?
     @State var logoutAlertShown = false
-    
+
     @ObservedObject var appState = AppState.shared
-    
+
     var body: some View {
         NavigationView {
             VStack {
@@ -53,7 +53,7 @@ struct DetailedCallingView: View {
                     Text("Call details")
                         .font(.title)
                         .padding()
-                    
+
                     HStack {
                         Spacer()
                         ZStack {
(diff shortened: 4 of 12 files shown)


DemoApp/Resources/Entitlements/DemoApp-Debug.entitlements
```diff
@@ -6,8 +6,7 @@
 	<string>development</string>
 	<key>com.apple.developer.associated-domains</key>
 	<array>
-		<string>applinks:staging.getstream.io</string>
```
ipavlidakis, line 9: This isn't needed as the domain is the same and can be handled.

DemoApp/Sources/Components/AppEnvironment.swift
```diff
@@ -229,7 +229,7 @@ extension AppEnvironment {
         case .debug:
             return [.staging, .pronto, .production]
         case .test:
-            return [.staging]
```
ipavlidakis, line 232: Update test environment to be able to process all available deeplinks

DemoApp/Sources/Components/MemoryLogDestination/OSLogDestination.swift
```diff
@@ -0,0 +1,14 @@
...
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamVideo
+import OSLog
+
+final class OSLogDestination: BaseLogDestination {
+
+    override func write(message: String) {
```
ipavlidakis, line 11: This allows us to see our logs from the AppStore versions using the Console.app

DemoApp/Sources/Views/CallView/CallingView/DetailedCallingView.swift
```diff
@@ -140,12 +140,16 @@ struct DetailedCallingView: View {
             self.callId = callId
             viewModel.joinCall(callType: .default, callId: callId)
         }
+        .onReceive(appState.$deeplinkInfo) { deeplinkInfo in
```
ipavlidakis, line 143: Added this one to align the DetailedCallingView with the SimpleView and allow us to perform the DeeplinkTests

SwiftUIDemoAppUITests/Tests/DeeplinkTests.swift
```diff
@@ -0,0 +1,147 @@
...
+        // We are launching and terminating the app to ensure the executable
+        // has been installed.
+        app.launch()
+        app.terminate()
+    }
+
+    override func setUpWithError() throws {
+        launchApp = false
+        try super.setUpWithError()
+    }
```
testableapple, line 24: Why do we need `launchApp = false` if we then anyways launch and terminate the app? The app should be installed without launching and terminating the app 🤔

SwiftUIDemoAppUITests/Tests/DeeplinkTests.swift
```diff
@@ -0,0 +1,147 @@
...
+            // synchronously waits for a button to be pressed the first time.
+            self.system.open(url)
+        }
+    }
+}
+
+enum Safari {
+
+    private static let safari = XCUIApplication(bundleIdentifier: "com.apple.mobilesafari")
+
```
testableapple, line 80: Can we move this to a separate file in `TestTools` for instance? Thit might be helpful for other test classes as well as for UIKit tests.

SwiftUIDemoAppUITests/Tests/DeeplinkTests.swift
```diff
@@ -0,0 +1,147 @@
...
+        let goButton = safari.buttons["Go"]
+        if goButton.waitForExistence(timeout: 5) {
+            safari.tapKeyboardKey("Go")
+        }
+    }
+
+    fileprivate static func open(_ url: URL) {
+        safari.launch()
+
+        _ = safari.wait(for: .runningForeground, timeout: 5)
```
testableapple, line 93: Does this mean that every test will have to wait for 5 sec? Any chance we could wait for some element to appear instead of sleeping for 5 sec?

DemoApp/Sources/Views/CallView/CallingView/DetailedCallingView.swift
```diff
@@ -107,7 +107,7 @@ struct DetailedCallingView: View {
             .alignedToReadableContentGuide()
         }
         .navigationViewStyle(StackNavigationViewStyle())
-//        .navigationBarHidden(true)
+        //        .navigationBarHidden(true)
```
martinmitrevski, line 110: Best to delete commented out code. 🙏

SwiftUIDemoAppUITests/Tests/DeeplinkTests.swift
```diff
@@ -0,0 +1,147 @@
+//
```
martinmitrevski, line 1: Stream header.

SwiftUIDemoAppUITests/Tests/DeeplinkTests.swift
```diff
@@ -0,0 +1,147 @@
...
+    fileprivate static func openUniversalLinkFromSmartBanner(_ url: URL) {
+        open(url)
+
+        let allowButton1 = safari.alerts.buttons["Allow"]
+        if allowButton1.waitForExistence(timeout: 5) {
+            allowButton1.tap()
+            let allowButton2 = safari.alerts.buttons["Allow"]
+            if allowButton2.waitForExistence(timeout: 5) {
+                allowButton2.tap()
+            }
```
testableapple, line 122: `safari.alerts.buttons["Allow"]` is just a locator, it does not change, I think there is no need to assign it multiple times 🤔

```suggestion
        let allowButton = safari.alerts.buttons["Allow"]
        if allowButton.waitForExistence(timeout: 5) {
            allowButton.tap()
            if allowButton.waitForExistence(timeout: 5) {
                allowButton.tap()
            }
```


By the way, if these `allow` buttons appear all the time, we can use `wait()`, something like this:

```suggestion
        let allowButton = safari.alerts.buttons["Allow"]
       allowButton.wait().tap()
       allowButton.wait().tap()
```

SwiftUIDemoAppUITests/Tests/DeeplinkTests.swift
```diff
@@ -0,0 +1,147 @@
+//
+//  DeeplinkTests.swift
+//  SwiftUIDemoAppUITests
+//
+//  Created by Ilias Pavlidakis on 25/9/23.
+//
+
+import XCTest
+
+final class DeeplinkTests: StreamTestCase {
```
martinmitrevski, line 10: Awesome 🚀  When are these tests being run? Should we put them as part of the checks or in a test plan?

SwiftUIDemoAppUITests/Tests/DeeplinkTests.swift
```diff
@@ -0,0 +1,147 @@
...
+
+    private static let safari = XCUIApplication(bundleIdentifier: "com.apple.mobilesafari")
+
+    fileprivate static func go(to url: URL) {
+        safari.textFields["Address"].tap()
+        safari.typeText(url.absoluteString)
+        let goButton = safari.buttons["Go"]
+        if goButton.waitForExistence(timeout: 5) {
+            safari.tapKeyboardKey("Go")
+        }
```
testableapple, line 87: Interesting, does it mean that tapping on `safari.buttons["Go"]` instead of `safari.tapKeyboardKey("Go")` does not work?

martinmitrevski (commented):
great stuff 👏  Just few small nits and we can 🚢

testableapple (commented):
Nicely done, @ipavlidakis!
Looks really great, I just have a couple of questions.

SwiftUIDemoAppUITests/Tests/DeeplinkTests.swift
ipavlidakis: We are only launching the app on the static setUp which is being called only once (just before any test case from the suite start running). In that way, we don't take the hit of that step every time but only once. 

> The app should be installed without launching and terminating the app
Amazingly it's not :(

SwiftUIDemoAppUITests/Tests/DeeplinkTests.swift
ipavlidakis: That's right! The keyboard introduction is not really behaving the same as the rest of the buttons

SwiftUIDemoAppUITests/Tests/DeeplinkTests.swift
ipavlidakis: They are part of the UI Tests so i believe they should be running as part of the Release. Not sure about the PR checks though. Maybe @testableapple can provide more insight to this one.

testableapple: It will be executed with the other tests

martinmitrevski (approved):
LGTM! ✅

PR #171, opened 2023-10-02
Do not generate a new Info.plist during the build

testableapple:


CallIntent/Info.plist
```diff
@@ -13,9 +13,9 @@
 	<key>CFBundleName</key>
 	<string>$(PRODUCT_NAME)</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.0.0</string>
```
ipavlidakis, line 16: We don't need those, right?

ipavlidakis (approved):

PR #173, opened 2023-10-04
Implement VisibilityThresholdModifier

ipavlidakis:
### 🎯 Goal

Reduce CPU usage demands during a call with more than 5 participants.

### 📝 Summary

When in a call with more than 5 participants we use a GridView to show and arrange video tracks. This view leverages the `onAppear` & `onDisappear` hooks to disable tracks that are not on screen. Unfortunately, as we are using `LazyVStack` to wrap our video tracks, iOS preloads views that aren't on screen, resulting in rendering video tracks that aren't really on user's viewport.

### 🛠 Implementation

To resolve this problem, we introduce a ViewModifier that internally uses a GeometryReader to compute if it's in its parent's viewport. If not the View is considered hidden and we propagate a visibility change that stops renderings for the View's video track.

### 🎨 Showcase

|  Before  |  After  |
| -------- | ------- |
|  image   |  image  |

#### Note
The smaller CPU Usage and Energy impact is because the app was unresponsive and render requests were dropping (as can be seen from the FPS counter).

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

- [x] Comparison screenshots added for visual changes


Sources/StreamVideoSwiftUI/CallView/CallView.swift
```diff
@@ -159,18 +159,11 @@ public struct CallView<Factory: ViewFactory>: View {
         viewFactory.makeVideoParticipantsView(
             viewModel: viewModel,
             availableSize: size,
-            onViewRendering: handleViewRendering(_:participant:),
```
martinmitrevski, line 162: Is it possible to do this without breaking changes to the public API?

Sources/StreamVideoSwiftUI/CallView/ParticipantsGridView.swift
```diff
@@ -8,37 +8,40 @@ import WebRTC
...
+        GeometryReader { geometryProxy in
+            ScrollView {
+                if #available(iOS 14.0, *) {
+                    LazyVGrid(
+                        columns: [
+                            .init(.adaptive(minimum: size.width, maximum: size.width), spacing: 0)
+                        ],
+                        spacing: 0
+                    ) {
+                        participantsContent(geometryProxy.frame(in: .global))
```
martinmitrevski, line 29: This approach also makes the usage a bit tied to our grid component. What about the spotlight and other views that customers might want to use in their apps?

Sources/StreamVideoSwiftUI/ViewFactory.swift
```diff
@@ -49,7 +49,6 @@ public protocol ViewFactory: AnyObject {
     func makeVideoParticipantsView(
         viewModel: CallViewModel,
         availableSize: CGSize,
-        onViewRendering: @escaping (VideoRenderer, CallParticipant) -> Void,
```
martinmitrevski, line 52: If we decide to change the public api, we should also update the docs and tutorials where this is used. But let's see if we can avoid that.

martinmitrevski (commented):
According to the screenshots, the device would heat more with the new approach, only the FPS is better, right?

ipavlidakis:
> According to the screenshots, the device would heat more with the new approach, only the FPS is better, right?

The energy impact isn't really representative for the thermal state. The device feels cooler with the latest changes. That doesn't mean that processing video isn't an energy impactful operation. The difference is that it takes more time to jump the thermal state levels.

Sources/StreamVideoSwiftUI/CallView/CallView.swift
ipavlidakis: I remove this one as it's not being used anywhere. We can definitely go ahead without this change but makes our API confusing as we are asking for something what we won't be using.

Sources/StreamVideoSwiftUI/CallView/ParticipantsGridView.swift
ipavlidakis: Spotlight also has this updated. What part of the approach seems to you that is tied to the Grid?

martinmitrevski: Where exactly is this modifier used in spotlight? I don't see it.

Sources/StreamVideo/WebRTC/StreamVideoCaptureHandler.swift
```diff
@@ -107,3 +111,7 @@ class StreamVideoCaptureHandler: NSObject, RTCVideoCapturerDelegate {
        )
     }
 }
+
+extension StreamVideoCaptureHandler: @unchecked Sendable {}
+extension RTCVideoFrame: @unchecked Sendable {}
```
martinmitrevski, line 116: better to have @preconcurrency import here.

Sources/StreamVideoSwiftUI/CallView/ParticipantsGridView.swift
```diff
@@ -8,37 +8,40 @@ import WebRTC
...
-                    ],
-                    spacing: 0
-                ) {
-                    participantsContent
-                }
-                .frame(width: availableSize.width)
-            } else {
-                VStack {
-                    participantsContent
+        GeometryReader { geometryProxy in
```
martinmitrevski, line 20: Seems like we might have an extra geometry reader? The `availableSize` value should also come from a geometry reader. Ideally, we should use only one.

Sources/StreamVideoSwiftUI/CallView/VisibilityThresholdModifier.swift
```diff
@@ -0,0 +1,107 @@
...
+                    let maxY = geometryInGlobal.maxY
+                    let minX = geometryInGlobal.minX
+                    let maxX = geometryInGlobal.maxX
+
+                    /// Calculate required height and width based on visibility threshold.
+                    let requiredHeight = geometry.size.height * threshold
+                    let requiredWidth = geometry.size.width * threshold
+
+                    /// Check if the content view is vertically within the parent's bounds.
+                    let verticalVisible = (minY + requiredHeight < bounds.maxY && minY > bounds.minY) ||
```
martinmitrevski, line 64: Can we write some tests for the calculations here?

StreamVideoTests/Utils/SerialActor_Tests.swift
```diff
@@ -0,0 +1,41 @@
+//
```
martinmitrevski, line 1: Stream header

docusaurus/docs/iOS/03-guides/10-view-slots.mdx
```diff
@@ -62,13 +62,11 @@ The video participants view slot presents the grid of users that are in the call
 public func makeVideoParticipantsView(
     participants: [CallParticipant],
     availableSize: CGSize,
-    onViewRendering: @escaping (VideoRenderer, CallParticipant) -> Void,
```
martinmitrevski, line 65: We don't use this in the tutorials, only in the docs?

StreamVideoTests/Utils/SerialActor_Tests.swift
```diff
@@ -0,0 +1,41 @@
...
+import XCTest
+@testable import StreamVideo
+
+final class SerialActor_Tests: XCTestCase {
+
+    private var count = 0
+
+    // Test to ensure tasks are executed serially.
+    func testSerialExecution() async throws {
+        let actor = SerialActor()
```
martinmitrevski, line 18: best to avoid keywords as variable names.

Sources/StreamVideo/WebRTC/StreamVideoCaptureHandler.swift
```diff
@@ -33,26 +35,28 @@ class StreamVideoCaptureHandler: NSObject, RTCVideoCapturerDelegate {
     }
     
     func capturer(_ capturer: RTCVideoCapturer, didCapture frame: RTCVideoFrame) {
-        Task { [weak self] in
-            guard let self else { return }
+        Task { [serialActor] in
+            await serialActor.enqueue { [weak self] in
```
martinmitrevski, line 39: Not 100% sure why enqueuing async tasks helps performance, can you please elaborate one more time? 🙏

Sources/StreamVideoSwiftUI/CallView/ParticipantsGridView.swift
ipavlidakis: `availableSize` is indeed coming from a GeometryReader in the CallView. We can't really use it here as we don't need only the size but the origin too. 

We can potentially use only one GeometryReader but that will require a change on the API that will allow us to pass `availableRect/availableViewport: CGRect` instead of `availableSize: CGSize`.

Sources/StreamVideoSwiftUI/CallView/VisibilityThresholdModifier.swift
ipavlidakis: 100%! It was my mistake to not add some from the beginning.

docusaurus/docs/iOS/03-guides/10-view-slots.mdx
ipavlidakis: It seems that the tutorial not include this.

Sources/StreamVideo/WebRTC/StreamVideoCaptureHandler.swift
ipavlidakis: The idea is that Tasks not running on a specific actor will run as part of the GlobalActor. The previous days I remember reading that the GlobalActor is concurrent. That meant that the Tasks created here (outside of an Actor) will run as part of the GlobalActor. As they will run concurrently we may doing processing for more that one frame at the same time which can be consuming.

However, looking again now I couldn't find what I was reading and i ended up asking ChatGPT. The answer is that the GlobalActor is serial. So I will remove this one.

Sources/StreamVideoSwiftUI/CallView/ParticipantsGridView.swift
martinmitrevski: I would say it's better to break the API at this point, since we are still in beta. Geometry readers can be expensive, best to use them only when needed. Wdyt?

ipavlidakis: Changing this one I think will incur many changes. How do you feel about having it in another PR?

martinmitrevski: Maybe another PR towards this PR, to see how much it is? Since if we do breaking changes, we should do them very soon.

ipavlidakis: A PR to this branch sounds good. Here it is #176

StreamVideoSwiftUITests/CallView/__Snapshots__/ParticipantsSpotlightLayout_Tests/test_spotlight_participantsConnectionQuality_snapshot.default-light-good.png
ipavlidakis, line 1: This one is higher as now we are using the frame from global and in the empty space is where we have the controls.

martinmitrevski (approved):
lgtm! ✅

PR #174, opened 2023-10-04
Adaptive frame rate depending on thermalState

ipavlidakis:
### 🎯 Goal

Adapt framerate for VideoRenderer (in oder to reduce CPU usage) when device's thermal state changes.

### 🛠 Implementation

By observing the thermal state we respond to changes on the thermal state. The goal is to reduce CPU requirements to render video tracks in order to allow the device to cool down or maintain its usage to a steady level.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideo/Utils/ThermalStateObserver.swift
+++ b/Sources/StreamVideo/Utils/ThermalStateObserver.swift
@@ -9,15 +9,31 @@ extension LogSubsystem {
     public static let thermalState = Self(rawValue: 1 << 6)
 }
 
+/// A protocol representing an observer of a device's thermal state.
+///
+/// This protocol is designed to provide both the current thermal state of the device and a publisher
+/// for monitoring changes to that state. Additionally, it provides a scaling factor to adapt the behavior
+/// or appearance of certain features based on the device's thermal condition.
 public protocol ThermalStateObserving: ObservableObject {
 
+    /// The current thermal state of the device.
     var state: ProcessInfo.ThermalState { get }
 
+    /// A publisher emitting updates when the thermal state changes.
     var statePublisher: AnyPublisher<ProcessInfo.ThermalState, Never> { get }
 
+    /// A scaling factor derived from the device's thermal state.
+    ///
+    /// This scale can be used to adapt functionalities, such as adjusting the resolution of streaming content,
+    /// to ensure optimal performance under varying thermal conditions.
     var scale: CGFloat { get }
 }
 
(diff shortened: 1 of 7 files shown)


Sources/StreamVideo/Utils/ThermalStateObserver.swift
```diff
@@ -9,11 +9,20 @@ extension LogSubsystem {
     public static let thermalState = Self(rawValue: 1 << 6)
 }
 
-public final class ThermalStateObserver: ObservableObject {
-    public static let shared = ThermalStateObserver()
+public protocol ThermalStateObserving: ObservableObject {
```
martinmitrevski, line 12: Since it's public API, best to add some Swift docs here.

Sources/StreamVideoSwiftUI/CallView/VideoRenderer.swift
```diff
@@ -93,21 +94,54 @@ public struct VideoRendererView: UIViewRepresentable {
...
+    private var cancellable: AnyCancellable?
+
+    private(set) var preferredFramesPerSecond: Int = UIScreen.main.maximumFramesPerSecond {
+        didSet {
+            metalView?.preferredFramesPerSecond = preferredFramesPerSecond
+            log.debug("🔄 preferredFramesPerSecond was updated to \(preferredFramesPerSecond).")
+        }
     }
-    
+    private lazy var metalView: MTKView? = { subviews.compactMap { $0 as? MTKView }.first }()
```
martinmitrevski, line 114: Why not self? How are we sure that `MTKView` is available? Previously we were setting the preferred FPS for `RTCMTLVideoView` (the current view).

ipavlidakis: We cannot use `self.preferredFramesPerSecond` as `RTCMTLVideoView` is a UIView and doesn't have this property available. If you look in the implementation of [RTCMTLVideoView](https://chromium.googlesource.com/external/webrtc/+/fc8c97f950a4fead4a5debe3c838942db1834010/webrtc/sdk/objc/Framework/Classes/Metal/RTCMTLVideoView.m#63) you will see that as long as metal is available the metalView is guaranteed to be there

martinmitrevski: 👍 I realized I actually had the same code written before, why am I complaining now 😅

martinmitrevski (approved):

PR #175, opened 2023-10-04
Use single instance for the microphoneChecker

ipavlidakis:
### 🎯 Goal

MicrophoneChecker's operations to startListening(create new recording file), update volume meters and stop recording (delete recording file) can be quite expensive when occuring multiple times per second. 

### 📝 Summary

We are using our injection system to create an instance for the MicrophoneChecker that can be overriden.

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/Sources/Components/DemoAppViewFactory.swift
+++ b/DemoApp/Sources/Components/DemoAppViewFactory.swift
@@ -12,7 +12,6 @@ final class DemoAppViewFactory: ViewFactory {
     static let shared = DemoAppViewFactory()
 
     @Injected(\.colors) var colors
-    @Injected(\.microphoneChecker) var microphoneChecker
 
     func makeWaitingLocalUserView(viewModel: CallViewModel) -> some View {
         DemoWaitingLocalUserView(viewFactory: self, viewModel: viewModel)
@@ -45,7 +44,6 @@ final class DemoAppViewFactory: ViewFactory {
     func makeCallView(viewModel: CallViewModel) -> DemoCallView<DemoAppViewFactory> {
         DemoCallView(
             viewFactory: self,
-            microphoneChecker: microphoneChecker,
             viewModel: viewModel
         )
     }
--- a/DemoApp/Sources/Components/DemoAppViewFactory.swift
+++ b/DemoApp/Sources/Components/DemoAppViewFactory.swift
@@ -12,6 +12,7 @@ final class DemoAppViewFactory: ViewFactory {
     static let shared = DemoAppViewFactory()
 
     @Injected(\.colors) var colors
+    @Injected(\.microphoneChecker) var microphoneChecker
 
     func makeWaitingLocalUserView(viewModel: CallViewModel) -> some View {
         DemoWaitingLocalUserView(viewFactory: self, viewModel: viewModel)
@@ -44,7 +45,7 @@ final class DemoAppViewFactory: ViewFactory {
     func makeCallView(viewModel: CallViewModel) -> DemoCallView<DemoAppViewFactory> {
         DemoCallView(
             viewFactory: self,
-            microphoneChecker: MicrophoneChecker(),
+            microphoneChecker: microphoneChecker,
             viewModel: viewModel
         )
     }
--- a/DemoApp/Sources/Views/CustomCallView.swift
+++ b/DemoApp/Sources/Views/CustomCallView.swift
@@ -9,11 +9,11 @@ import SwiftUI
 struct CustomCallView<Factory: ViewFactory>: View {
     
     @Injected(\.colors) var colors
+    @Injected(\.microphoneChecker) var microphoneChecker
     
     var viewFactory: Factory
     @ObservedObject var viewModel: CallViewModel
-    
-    @StateObject var microphoneChecker = MicrophoneChecker()
+
     @State var mutedIndicatorShown = false
     
     var body: some View {
--- a/DemoApp/Sources/Views/CustomCallView.swift
+++ b/DemoApp/Sources/Views/CustomCallView.swift
@@ -12,8 +12,8 @@ struct CustomCallView<Factory: ViewFactory>: View {
 
     var viewFactory: Factory
     @ObservedObject var viewModel: CallViewModel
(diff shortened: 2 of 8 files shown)


Sources/StreamVideoSwiftUI/CallingViews/MicrophoneChecker.swift
```diff
@@ -173,3 +173,20 @@ protocol AudioSessionProtocol {
...
 extension AVAudioSession: AudioSessionProtocol {}
+
+/// Provides the default value of the `ThermalStateObserving` protocol.
+public struct MicrophoneCheckerKey: InjectionKey {
+    public static var currentValue: MicrophoneChecker = MicrophoneChecker()
+}
+
+extension InjectedValues {
+
+    public var microphoneChecker: MicrophoneChecker {
```
martinmitrevski, line 184: Not sure if this should be public? Any strong arguments for it?

ipavlidakis: I don't have strong arguments! The less public the better!

ipavlidakis: On a second thought, if integrators would like to add more valueLimits, they will need to override the views and provide a different instance. By having this public they will be able to inject a differently configured instance that will be used from SDK's views without further changes. For this reason it's probably better to keep it as public.

martinmitrevski: It actually seems to be already used in the demo app, which prevents building the app: 

❌  /Users/runner/work/stream-video-swift/stream-video-swift/DemoApp/Sources/Components/DemoAppViewFactory.swift:15:15: key path value type 'WritableKeyPath<InjectedValues, MicrophoneChecker>' cannot be converted to contextual type 'KeyPath<InjectedValues, MicrophoneChecker>'

    @Injected(\.microphoneChecker) var microphoneChecker
        ^



❌  /Users/runner/work/stream-video-swift/stream-video-swift/DemoApp/Sources/Views/CustomCallView.swift:12:15: 'microphoneChecker' is inaccessible due to 'internal' protection level

    @Injected(\.microphoneChecker) var microphoneChecker
              ^



❌  /Users/runner/work/stream-video-swift/stream-video-swift/DemoApp/Sources/Views/CustomCallView.swift:12:15: key path value type 'WritableKeyPath<InjectedValues, MicrophoneChecker>' cannot be converted to contextual type 'KeyPath<InjectedValues, MicrophoneChecker>'

Sources/StreamVideoSwiftUI/CallView/LocalParticipantViewModifier.swift
```diff
@@ -10,16 +10,18 @@ import StreamVideo
 public struct LocalParticipantViewModifier: ViewModifier {
 
     private let localParticipant: CallParticipant
-    @StateObject private var microphoneChecker: MicrophoneChecker
+    @Injected(\.microphoneChecker) var microphoneChecker
```
martinmitrevski, line 13: I think if you do:
```swift
@ObservedObject var microphoneChecker = InjectedValues[\.microphoneChecker]
```
It would simplify the code a bit. This is also ok, but would be nice to avoid storing extra state.

Sources/StreamVideoSwiftUI/CallingViews/iOS13/PreJoiningView_iOS13.swift
```diff
@@ -11,7 +11,7 @@ struct LobbyView_iOS13: View {
     
     @ObservedObject var callViewModel: CallViewModel
     @BackportStateObject var viewModel: LobbyViewModel
-    @BackportStateObject var microphoneChecker = MicrophoneChecker()
+    @Injected(\.microphoneChecker) var microphoneChecker: MicrophoneChecker
```
martinmitrevski, line 14: hm, did you test on iOS 13 to make sure it still works there?

ipavlidakis: It has actually become quite difficult to test iOS 13 on Xcode 14/15 (i haven't tried it in a while). The code structure is now as it used to be but it may take a while to be able to test it.

DemoApp/Sources/Views/CustomCallView.swift
```diff
@@ -9,11 +9,11 @@ import SwiftUI
...
     
     @Injected(\.colors) var colors
-    
+
     var viewFactory: Factory
     @ObservedObject var viewModel: CallViewModel
-    
-    @StateObject var microphoneChecker = MicrophoneChecker()
+
+    @StateObject var microphoneChecker = InjectedValues[\.microphoneChecker]
```
martinmitrevski, line 16: Should actually be an `@ObservedObject` in this case, since the view doesn't own it anymore.

Sources/StreamVideoSwiftUI/CallingViews/iOS13/PreJoiningView_iOS13.swift
```diff
@@ -11,8 +11,8 @@ struct LobbyView_iOS13: View {
     
     @ObservedObject var callViewModel: CallViewModel
     @BackportStateObject var viewModel: LobbyViewModel
-    @BackportStateObject var microphoneChecker = MicrophoneChecker()
-    
+    @BackportStateObject var microphoneManager = InjectedValues[\.microphoneManager]
```
martinmitrevski, line 14: Here and in few other places as well.

DemoApp/Sources/Components/DemoAppViewFactory.swift
```diff
@@ -12,6 +12,7 @@ final class DemoAppViewFactory: ViewFactory {
     static let shared = DemoAppViewFactory()
 
     @Injected(\.colors) var colors
+    @Injected(\.microphoneChecker) var microphoneChecker
```
martinmitrevski, line 15: I think we don't need it here anymore, right?

martinmitrevski (approved):

PR #181, opened 2023-10-10
Fix ChatView disappearing when call events happening

ipavlidakis:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/577

### 📝 Summary

ChatView was disappearing when call events where happening. 

### 🛠 Implementation

The issue is that the ChatViewModel we rely on wasn't hold as StateObject somewhere, which causing the creation of multiple instances that were then affecting the ChatView presentation.

### 🧪 Manual Testing Notes

- Join a call
- Open chat
- Start sharing from another paricipant

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/Sources/Views/CallView/DemoChatModifier.swift
+++ b/DemoApp/Sources/Views/CallView/DemoChatModifier.swift
@@ -1,0 +1,47 @@
+//
+//  DemoChatModifier.swift
+//  DemoApp
+//
+//  Created by Ilias Pavlidakis on 10/10/23.
+//
+
+import Foundation
+import SwiftUI
+import StreamVideoSwiftUI
+
+struct ChatModifier: ViewModifier {
+
+    @ObservedObject var viewModel: CallViewModel
+    @ObservedObject var chatViewModel: DemoChatViewModel
+
+    func body(content: Content) -> some View {
+        content
+            .halfSheetIfAvailable(
+                isPresented: $chatViewModel.isChatVisible,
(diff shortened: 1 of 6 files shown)


DemoApp/Sources/Views/CallView/DemoChatModifier.swift
```diff
@@ -0,0 +1,47 @@
+//
```
martinmitrevski, line 1: Stream header 😅

DemoApp/Sources/Views/CallView/DemoChatModifier.swift
```diff
@@ -0,0 +1,47 @@
...
+//  DemoApp
+//
+//  Created by Ilias Pavlidakis on 10/10/23.
+//
+
+import Foundation
+import SwiftUI
+import StreamVideoSwiftUI
+
+struct ChatModifier: ViewModifier {
```
martinmitrevski, line 12: I like this idea of having chat modifier. Maybe we could introduce chat extensions for video as reusable components in the future 👍

martinmitrevski (approved):
lgtm ✅ just one header needs fixing

DemoApp/Sources/Views/CallView/DemoChatModifier.swift
ipavlidakis: Sounds a good idea and will make integration much easier for customers. I think we should have that in our roadmap

PR #182, opened 2023-10-10
Provide calls in the demo app

martinmitrevski:
### 🎯 Goal

Allow Stream employees to have calls with the demo app.

### 🛠 Implementation

Usage of the GoogleSignIn and Contacts API. 

### 🎨 Showcase

|  Before  |  After  |
| -------- | ------- |
|  `img`   |  `img`  |

### 🧪 Manual Testing Notes

Login with your GetStream account to test this.

### ☑️ Contributor Checklist

### 🎁 Meme

ALT_TEXT


DemoApp/Info.plist
```diff
@@ -36,9 +36,21 @@
 				<string>streamvideo</string>
 			</array>
 		</dict>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLName</key>
+			<string>com.googleusercontent.apps.347024607410-48j4atipav0tcr4pesap4elr1u9t11uh</string>
```
ipavlidakis, line 43: Probably this one should be hidden from public, right? We can use build variables and inject it when releasing a new build

DemoApp/Sources/Components/Router.swift
```diff
@@ -76,9 +77,13 @@ final class Router: ObservableObject {
         if AppEnvironment.configuration == .test, AppEnvironment.contains(.mockJWT) {
             return
         } else if let userCredentials = AppState.shared.unsecureRepository.loadCurrentUser() {
-            appState.currentUser = userCredentials.userInfo
-            appState.userState = .loggedIn
-            handleLoggedInUserCredentials(userCredentials, deeplinkInfo: .empty)
+            if userCredentials.userInfo.id.contains("@getstream") {
+                GIDSignIn.sharedInstance.restorePreviousSignIn { [weak self] _,_ in
+                    self?.setupUser(with: userCredentials)
```
ipavlidakis, line 82: Is the an error being returned here that should be handling?

DemoApp/Sources/Views/CallView/DemoCallingTopView.swift
```diff
@@ -40,6 +42,14 @@ struct DemoCallingTopView: View {
             }
 
             Spacer()
+            
+            if GIDSignIn.sharedInstance.currentUser != nil {
```
ipavlidakis, line 46: Should we make that available only on pre-release builds?

DemoApp/Sources/Views/Login/LoginView.swift
```diff
@@ -53,6 +53,17 @@ struct LoginView: View {
                     } icon: {
                         Image(systemName: "person.crop.circle.badge.clock.fill")
                     }
+                    
+                    LoginItemView {
```
ipavlidakis, line 57: Should we hide this one in Release builds?

DemoApp/Sources/Components/Router.swift
martinmitrevski: No, if a google sign in fails to happen, we just setup the user as normal. For example, your id can contain @getstream, but it's not done via Google sign in.

DemoApp/Sources/Views/CallView/DemoCallingTopView.swift
martinmitrevski: I think we should have it on the release as well. The sign in fails if you don't have getstream account. Would probably encourage employees to use it more, while there's no impact for other users.

DemoApp/Sources/Views/Login/LoginView.swift
martinmitrevski: Same as above.

DemoApp/Info.plist
martinmitrevski: How do we do that? Do we have an example to follow?

DemoApp/Sources/Views/CallView/DemoCallingTopView.swift
DanielNovak: On Android production builds (in Google Play) we don't have a manul sign-in flow yet, but maybe we could add a button to login with Google (which would log out the random user and log you in with Google).

martinmitrevski: I think it's important that it's not easily visible for a non-Stream user (i.e. it's not on the main page). That's why we put it in the login, which I think it's a good place.
Other option is to invoke it with some hidden gesture maybe?

DemoApp/Info.plist
ipavlidakis: We can have it as a build variable in CI environment (in the same way we are using some credentials for testing). @testableapple  can help probably

DemoApp/Sources/Views/CallView/DemoCallingTopView.swift
ipavlidakis: The login page isn't visible or reachable in Release builds. That's why i think that we should be hiding it for non-internal builds otherwise we need to move it to a screen that's reachable or make the login screen reachable

fastlane/Fastfile
```diff
@@ -16,6 +16,8 @@ app_secret = ENV.fetch('STREAM_VIDEO_SECRET', nil)
 derived_data_path = 'derived_data'
 source_packages_path = 'spm_cache'
 buildcache_xcargs = 'CC=clang CPLUSPLUS=clang++ LD=clang LDPLUSPLUS=clang++'
+gci = 'com.googleusercontent.apps.347024607410-48j4atipav0tcr4pesap4elr1u9t11uh'
```
ipavlidakis, line 19: As this is also in public doesn't it defy all the actions we took to hide it?. Can we use the environment variable?

Note: This is the reversed id. The regular one begins with the numbers

ipavlidakis (approved):

PR #184, opened 2023-10-10
Fix waiting user view appearance

ipavlidakis:
### 🧪 Manual Testing Notes

- Join a call where you will be the only participant
- The view to share the call's link should be visible
- Open the chat and close it
- The view should be hidden and then become visible again

### ☑️ Contributor Checklist

### 🎁 Meme

### DIFF
--- a/DemoApp/Sources/Views/WaitingLocalUserView/DemoWaitingLocalUserView.swift
+++ b/DemoApp/Sources/Views/WaitingLocalUserView/DemoWaitingLocalUserView.swift
@@ -14,6 +14,7 @@ struct DemoWaitingLocalUserView<Factory: DemoAppViewFactory>: View {
     @ObservedObject var viewModel: CallViewModel
 
     @State private var isSharePresented = false
+    @State private var isChatVisible = false
 
     private let viewFactory: Factory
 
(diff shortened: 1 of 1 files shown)


DemoApp/Sources/Views/WaitingLocalUserView/DemoWaitingLocalUserView.swift
```diff
@@ -30,7 +31,7 @@ struct DemoWaitingLocalUserView<Factory: DemoAppViewFactory>: View {
             viewFactory
                 .makeInnerWaitingLocalUserView(viewModel: viewModel)
 
-            if chatViewModel?.isChatVisible == true {
```
martinmitrevski, line 33: hmm, I don't think you need additional State for this - you already have the same one in place in the view model.
I notice that the check is different now (`== false` instead of `== true`). Could that be the fix instead?

ipavlidakis: The injected chatViewModel isn't posting updates, which means that this won't change when the chat presents/hides. 

I think previously we had a mistake and lately the popup isn't visible when the chat is closed.

martinmitrevski: why not make it `@ObservedObject`?
```swift
@ObservedObject var chatViewModel = InjectedValues[\.chatViewModel]!
```

ipavlidakis: Want to avoid explicitely unwrapping. Do you see any issues with the private @State?

PR #186, opened 2023-10-11
Align participants sorting with Web

ipavlidakis:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/582

### 🎯 Goal

Avoid moving video tracks when user properties change, especially when that user is currently visible.

### 📝 Summary

Based on React implementation [here](https://github.com/GetStream/stream-video-js/blob/main/packages/client/src/sorting/presets.ts)

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideo/CallState.swift
+++ b/Sources/StreamVideo/CallState.swift
@@ -301,32 +301,53 @@ public class CallState: ObservableObject {
         self.ownCapabilities = event.ownCapabilities
     }
 
-    private func didUpdate(_ participants: [CallParticipant]) {
-        self.participants = participants.sorted(using: defaultComparators)
+    private func didUpdate(_ newParticipants: [CallParticipant]) {
+        // Combine existing and newly added participants.
+        let currentParticipantIds = Set(self.participants.map(\.id))
+        let newlyAddedParticipants = Set(newParticipants.map(\.id))
+            .subtracting(currentParticipantIds)
+            .compactMap { participantsMap[$0] }
+
+        // Sort the updated participants.
+        let updatedCurrentParticipants: [CallParticipant] = (
+            self
+            .participants
+            .compactMap { participantsMap[$0.id] } + newlyAddedParticipants
+        )
+        .sorted(by: defaultComparators)
+
+        // Variables to hold segregated participants.
         var remoteParticipants: [CallParticipant] = []
         var activeSpeakers: [CallParticipant] = []
         var screenSharingSession: ScreenSharingSession?
 
-        for participant in participants {
+        // Segregate participants based on conditions.
+        for participant in updatedCurrentParticipants {
+            // Check if participant is local or remote.
             if participant.sessionId == sessionId {
                 localParticipant = participant
             } else {
                 remoteParticipants.append(participant)
             }
 
+            // Check if participant is speaking.
             if participant.isSpeaking {
                 activeSpeakers.append(participant)
             }
 
+            // Check if participant is a dominant speaker.
             if participant.isDominantSpeaker {
                 dominantSpeaker = participant
             }
 
+            // Check if participant is sharing their screen.
             if let screenshareTrack = participant.screenshareTrack, participant.isScreensharing {
                 screenSharingSession = .init(track: screenshareTrack, participant: participant)
             }
         }
 
+        // Update the respective class properties.
+        self.participants = updatedCurrentParticipants
         self.screenSharingSession = screenSharingSession
         self.remoteParticipants = remoteParticipants
         self.activeSpeakers = activeSpeakers
--- a/Sources/StreamVideo/Utils/Sorting.swift
+++ b/Sources/StreamVideo/Utils/Sorting.swift
@@ -20,7 +20,9 @@ public let defaultComparators: [StreamSortComparator<CallParticipant>] = [
     dominantSpeaker,
     ifInvisible(isSpeaking),
     ifInvisible(publishingVideo),
-    ifInvisible(publishingAudio)
+    ifInvisible(publishingAudio),
+    joinedAt,
+    userId
 ]
 
 /// The set of comparators used for sorting `CallParticipant` objects during livestreams.
@@ -34,13 +36,15 @@ public let livestreamComparators: [StreamSortComparator<CallParticipant>] = [
     ifInvisible(isSpeaking),
     ifInvisible(publishingVideo),
     ifInvisible(publishingAudio),
-    roles()
+    roles(),
+    joinedAt,
+    userId
 ]
 
 // MARK: - Sort Sequence
 
 /// Represents the possible orders for sorting.
-public enum SortOrder { case ascending, descending }
+public enum StreamSortOrder { case ascending, descending }
 
 extension Sequence where Element == CallParticipant {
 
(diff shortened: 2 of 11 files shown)


Sources/StreamVideo/Utils/Sorting.swift
```diff
@@ -4,110 +4,230 @@
...
+
+/// Comparator which sorts participants by the fact that they are speaking or not.
+public var isSpeaking: StreamSortComparator<CallParticipant> = { comparison($0, $1, keyPath: \.isSpeaking) }
+
+/// Comparator which prioritizes participants who are pinned.
+public var pinned: StreamSortComparator<CallParticipant> = { a, b in
+    switch (a.pin, b.pin) {
+    case (nil, _?): return .orderedDescending
+    case (_?, nil): return .orderedAscending
+    case (let aPin?, let bPin?) where !aPin.isLocal && bPin.isLocal: return .orderedAscending
```
martinmitrevski, line 190: shouldn't this be the opposite? Remote pins have priority over local ones.

Sources/StreamVideoSwiftUI/CallViewModel.swift
```diff
@@ -177,21 +174,8 @@ open class CallViewModel: ObservableObject {
         )
     }
     
-    public var participants: [CallParticipant] {
-        callParticipants
-            .filter {
-                if participantsLayout == .grid,
-                   callParticipants.count <= 3,
-                   (call?.state.screenSharingSession == nil || call?.state.isCurrentUserScreensharing == true) {
```
martinmitrevski, line 185: Why do we remove this condition? If I remember correctly, this prevented mirroring when the current user is screensharing. There are several cases here that we should check.

Sources/StreamVideoSwiftUI/CallViewModel.swift
```diff
@@ -380,24 +364,6 @@ open class CallViewModel: ObservableObject {
     ///  - participant: the participant whose track visibility would be changed.
     ///  - isVisible: whether the track should be visible.
     public func changeTrackVisibility(for participant: CallParticipant, isVisible: Bool) {
-        if !isVisible {
```
martinmitrevski, line 383: While some of this stuff might not be needed anymore, with the introduction of the visibility modifier, we need to test all cases. This code was here because of tracks being hidden when changing layouts. Did you test that everything works as expected?

StreamVideoTests/CallState/CallState_Tests.swift
```diff
@@ -0,0 +1,230 @@
...
+            expectedTransformer: { updated in
+                return [updated[2], updated[1], updated[0], updated[3]]
+            }
+        )
+
+    }
+
+    // MARK: - Private helpers
+
+    private func assertParticipantsUpdate(
```
martinmitrevski, line 206: This is very cool 👏

StreamVideoTests/Mock/CallParticipant_Mock.swift
```diff
@@ -0,0 +1,58 @@
...
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamVideo
+import WebRTC
+
+extension CallParticipant {
+
+    static func dummy(
```
martinmitrevski, line 11: It seems we have `makeCallParticipant` in the `MockResponseBuilder` that does pretty much the same thing. We should maybe align them, doesn't have to be in this PR.

StreamVideoTests/Utils/Sorting_Tests.swift
```diff
@@ -4,182 +4,741 @@
...
+                .dummy(),
+                .dummy(pin: PinInfo(isLocal: true, pinnedAt: Date(timeIntervalSince1970: 100)))
+            ],
+            comparator: pinned,
+            expectedTransformer: { [$0[1], $0[0]] }
+        )
+    }
+
+    /// Test with pinned participants having different pinnedAt dates.
+    func test_pinned_differentPinnedDates() {
```
martinmitrevski, line 50: I don't see a test with local and remote pin. That would've caught the issue with the wrong comparator.

martinmitrevski (commented):
looks great in general. However, we need to test few things before merging, since there's some logic that's removed that might have an impact. @testableapple can also help.

Sources/StreamVideo/CallState.swift
```diff
@@ -301,32 +301,53 @@ public class CallState: ObservableObject {
         self.ownCapabilities = event.ownCapabilities
     }
 
-    private func didUpdate(_ participants: [CallParticipant]) {
-        self.participants = participants.sorted(using: defaultComparators)
+    private func didUpdate(_ newParticipants: [CallParticipant]) {
+        // Combine existing and newly added participants.
+        let currentParticipantIds = Set(self.participants.map(\.id))
```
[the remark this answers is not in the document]
ipavlidakis: I have added a benchmark test to validate this one

Sources/StreamVideo/Utils/Sorting.swift
```diff
@@ -4,110 +4,230 @@
...
+            return comparator(a, b)
         }
     }
-    return .orderedSame
 }
 
-public var userId: StreamSortComparator<CallParticipant> = { (p1, p2) in
-    p1.id >= p2.id ? .orderedDescending : .orderedAscending
+/// Returns a comparator that always deems two elements "equal" regardless of their actual values.
+public func noop<T>() -> StreamSortComparator<T> { { _, _ in .orderedSame } }
```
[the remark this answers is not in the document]
ipavlidakis: We don't really. I just had it initially as React is using it but i didn't end up using it.

Sources/StreamVideo/Utils/Sorting.swift
ipavlidakis: You are right! I updated it and added the missing test case to validate it.

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: Can you elaborate more about what this purposed for? What do you mean by mirroring? I removed it to maintain consistency with state's participants and because screensharing is being handled from sorting. What shall i be looking for here?

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: Tests so far haven't yield anything concerning but i'll run another round with this one and the comment above just to be sure.

StreamVideoTests/Mock/CallParticipant_Mock.swift
ipavlidakis: Done!

ipavlidakis: Done!

StreamVideoTests/Utils/Sorting_Tests.swift
ipavlidakis: I just added one.

Sources/StreamVideoSwiftUI/CallViewModel.swift
martinmitrevski: I had one case where one track never got enabled. Hard to tell if it was us or the SFU.

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: This one is now covered by unit tests

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: That was related to this #191

martinmitrevski (approved):
great work 👏  I did some tests, but would be good if @testableapple does some more.

PR #235, opened 2023-11-21
Implemented fast reconnection

martinmitrevski:
### 🎯 Goal

Implement fast reconnect on iOS. Fast reconnect means you try to re-join without leaving the call. Other users will notice a brief freeze of you.

Check the spec on notion for more details.

### 🛠 Implementation

Mostly updated the WebRTC client.

### 🎨 Showcase

|  Before  |  After  |
| -------- | ------- |
|  `img`   |  `img`  |

### 🧪 Manual Testing Notes

It's a bit tricky to test, and the success of fast reconnect depends on many factors.

However, what's important is - even if fast reconnect fails, there's a full reconnect as a fallback.

Testing steps:
- join a call on WiFi
- quickly switch to 5g / other Wifi
- you should reconnect to the call without leaving the call

### ☑️ Contributor Checklist

### 🎁 Meme

ALT_TEXT

### DIFF
--- a/Sources/StreamVideo/WebRTC/WebRTCClient.swift
+++ b/Sources/StreamVideo/WebRTC/WebRTCClient.swift
@@ -167,6 +167,7 @@ class WebRTCClient: NSObject, @unchecked Sendable {
     private(set) var videoOptions = VideoOptions()
     private let environment: WebSocketClient.Environment
     private let apiKey: String
+    private let fastReconnectTimeout: TimeInterval = 4.0
     
     private var migratingSignalService: Stream_Video_Sfu_Signal_SignalServer?
     private var migratingWSClient: WebSocketClient?
@@ -174,7 +175,8 @@ class WebRTCClient: NSObject, @unchecked Sendable {
     private var fromSfuName: String?
     private var tempSubscriber: PeerConnection?
     private var currentScreenhsareType: ScreensharingType?
-
+    private var isFastReconnecting = false
+    
     @Injected(\.thermalStateObserver) private var thermalStateObserver
 
     var onParticipantsUpdated: (([String: CallParticipant]) -> Void)?
@@ -245,16 +247,19 @@ class WebRTCClient: NSObject, @unchecked Sendable {
         }
         addOnParticipantsChangeHandler()
         subscribeToAppLifecycleChanges()
+        subscribeToInternetConnectionUpdates()
     }
     
     func connect(
         callSettings: CallSettings,
         videoOptions: VideoOptions,
         connectOptions: ConnectOptions,
-        migrating: Bool = false
+        migrating: Bool = false,
+        fastReconnect: Bool = false
     ) async throws {
         let connectionStatus = await state.connectionState
-        if (connectionStatus == .connected || connectionStatus == .connecting) && !migrating {
+        let isReconnection = migrating || fastReconnect
+        if (connectionStatus == .connected || connectionStatus == .connecting) && !isReconnection {
             log.debug("Skipping connection, already connected or connecting", subsystems: .webRTC)
             return
         }
@@ -264,16 +269,21 @@ class WebRTCClient: NSObject, @unchecked Sendable {
         log.debug("Connecting to SFU", subsystems: .webRTC)
         await state.update(connectionState: .connecting)
         log.debug("Setting user media", subsystems: .webRTC)
-        if !migrating {
+        if !isReconnection {
+            isFastReconnecting = false
             await setupUserMedia(callSettings: callSettings)
             log.debug("Connecting WS channel", subsystems: .webRTC)
             signalChannel?.connect()
             sfuMiddleware.onSocketConnected = handleOnSocketConnected
-        } else {
+        } else if migrating {
             log.debug("Performing session migration", subsystems: .webRTC)
             migratingWSClient?.connect()
             publisher?.update(configuration: connectOptions.rtcConfiguration)
             sfuMiddleware.onSocketConnected = handleOnMigrationJoinResponse
+        } else if fastReconnect {
+            log.debug("Performing fast reconnect", subsystems: .webRTC)
+            signalChannel?.connect()
+            sfuMiddleware.onSocketConnected = handleOnSocketConnected
         }
         sfuMiddleware.onParticipantCountUpdated = { [weak self] participantCount in
             self?.onParticipantCountUpdated?(participantCount)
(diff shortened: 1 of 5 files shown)


Sources/StreamVideo/WebRTC/WebRTCClient.swift
```diff
@@ -1163,10 +1201,93 @@ class WebRTCClient: NSObject, @unchecked Sendable {
...
+            self.isFastReconnecting = false
+            let reconnectPublisher = publisher != nil ? publisher?.connectionState != .connected : false
+            let reconnectSubscriber = subscriber != nil ? subscriber?.connectionState != .connected : false
+            let shouldFullyReconnect = reconnectPublisher || reconnectSubscriber
+            if shouldFullyReconnect {
+                log.debug("Fast reconnect failed, doing full reconnect")
+                onSignalConnectionStateChange?(.disconnected(source: .noPongReceived))
+            } else {
+                log.debug("Fast reconnect successfull")
+            }
```
ipavlidakis, line 1273:
```suggestion
self.isFastReconnecting = false

func isPeerConnectionDisconnected(_ peerConnection: PeerConnection?) -> Bool {
    guard let peerConnection else {
        return true
    }

    switch peerConnection.connectionState {
    case .disconnected, .failed:
        return true
    default:
        return false
    }
}

let reconnectPublisher = isPeerConnectionDisconnected(publisher)
let reconnectSubscriber = isPeerConnectionDisconnected(subscriber)

let shouldFullyReconnect = reconnectPublisher || reconnectSubscriber
if shouldFullyReconnect {
    log.debug("Fast reconnect failed, doing full reconnect")
    onSignalConnectionStateChange?(.disconnected(source: .noPongReceived))
} else {
    log.debug("Fast reconnect successfully")
}
```

Android seems to considers that a connection needs to be reconnected if its state is `failed` or `disconnected`. ([reference](https://www.notion.so/stream-wiki/Reconnect-96bb776a1c2140ae9a7306dda4318f26?pvs=4#99eb4d9060144aaeb37ab5d717bd90e4))

Sources/StreamVideo/WebRTC/WebRTCClient.swift
```diff
@@ -1163,10 +1201,93 @@ class WebRTCClient: NSObject, @unchecked Sendable {
...
+    }
+    
+    @objc private func handleConnectionStateChange(_ notification: NSNotification) {
+        guard let status = notification.userInfo?[Notification.internetConnectionStatusUserInfoKey] as? InternetConnection.Status else {
+            return
+        }
+        
+        if status == .unavailable {
+            if !isFastReconnecting {
+                isFastReconnecting = true
```
ipavlidakis, line 1230: Here we should probably capture the current timestamp. Once we are back online, we can check the timestamp we captured and compare it with the current timestamp. If the diff is less or equal to the `fastReconnectTimeout` and the connections haven't been disconnected, then we should try the fast reconnect.

TBH the current approach will still work as we rely on the SFU to tell us if we can reconnect, but my comments are align the implementation with Android (and potentially React)

Sources/StreamVideo/WebRTC/WebRTCClient.swift
```diff
@@ -1163,10 +1201,93 @@ class WebRTCClient: NSObject, @unchecked Sendable {
...
+            }
+            return
+        }
+        if status.isAvailable && isFastReconnecting, let url = signalChannel?.connectURL {
+            signalChannel = makeWebSocketClient(
+                url: url,
+                apiKey: .init(apiKey),
+                isFastReconnect: true
+            )
```
ipavlidakis, line 1241: Do we need to update the signalChannel instance? Isn't there a way to request it to reconnect?

ipavlidakis (commented):
There a few more points from the Spec that Android is handling and seems that we don't.
1. They seem to be handling differently the WS disconnection and the Internet connection dropping ([reference](https://www.notion.so/stream-wiki/Reconnect-96bb776a1c2140ae9a7306dda4318f26?pvs=4#7ada78911f584a70b0112035ebdd59c6))
2. When trying to fully reconnect they are trying for up to 30 seconds and then they fail. Shall we also do that (probably as part of another PR)?

I tried connecting to another wifi (the guest network of my main one) but didn't manage to get the onto the fast reconnect path

Sources/StreamVideo/WebRTC/WebRTCClient.swift
martinmitrevski: Yes, that's a good point, already updated the PR.

Sources/StreamVideo/WebRTC/WebRTCClient.swift
```diff
@@ -1163,10 +1201,93 @@ class WebRTCClient: NSObject, @unchecked Sendable {
...
+    }
 }
 
 extension WebRTCClient: ConnectionStateDelegate {
     func webSocketClient(_ client: WebSocketClient, didUpdateConnectionState state: WebSocketConnectionState) {
-        onSignalConnectionStateChange?(state)
+        log.debug("WS connection state changed to \(state)")
+        switch state {
+        case .disconnected(source: _), .disconnecting(source: _):
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.3, execute: { [weak self] in
```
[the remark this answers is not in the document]
martinmitrevski: sometimes this method is called before the internet connection monitor, which is strange. This makes sure that we always do the checks there before taking action here.

Sources/StreamVideo/WebRTC/WebRTCClient.swift
martinmitrevski: thanks, this is cleaner, will update it. 👍

Sources/StreamVideo/WebRTC/WebRTCClient.swift
martinmitrevski: When you switch networks, your IP address changes. So all the TCP connections that the device has breaks. So when you do a fast reconnect, you have to re-establish a new WebSocket connection. 

This is because a TCP connection is identified by a 4-tuple which is (sourceIP, sourcePort, destinationIP, destinationPort). On network change, source IP and port change and the TCP connection is no longer valid.

Sources/StreamVideo/WebRTC/WebRTCClient.swift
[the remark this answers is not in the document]
ipavlidakis: nit: Can we move some where higher and give it a descriptive name (and ideally a comment) rather than using it down here?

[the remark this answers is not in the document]
martinmitrevski: Added a comment. What do you mean by putting it higher? This is an extension that implements the WS delegate.

Sources/StreamVideo/WebRTC/WebRTCClient.swift
```diff
@@ -1163,10 +1209,119 @@ class WebRTCClient: NSObject, @unchecked Sendable {
...
+                    connectOptions: connectOptions!,
+                    fastReconnect: true
+                )
+                checkFastReconnectionStatus()
+            }
+        }
+    }
+    
+    private func checkFastReconnectionStatus(retries: Int = 0) {
+        DispatchQueue.main.asyncAfter(deadline: .now() + fastReconnectTimeout) { [weak self] in
```
ipavlidakis, line 1262: Why do we still need to do the async after, now that we check the diff between connection and disconnection times?

Sources/StreamVideo/WebRTC/WebRTCClient.swift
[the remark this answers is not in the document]
martinmitrevski: Good point, used our `Constants` for that.

Sources/StreamVideo/WebRTC/WebRTCClient.swift
martinmitrevski: just in case, if the connection failed after the timeout, to start the full reconnect.

testableapple (approved):

ipavlidakis (approved):

PR #251, opened 2023-12-11
[Fix]VideoRenderers being added in multiple superviews

ipavlidakis:
### 🎯 Goal

Ensure that video tracks are always visible.

### 📝 Summary

There are a two underline issues resulting in this bug:
- When a user stops publishing video (that is translated as track muted for BE) we remove the track from our array and then we stop receiving events about it whenever the user decides to re-enable their camera.
- The visibility observation was being added on the view ONLY when the participant had video. When the user reenables their video the view doesn't fire the geometry calculation causing the view to not trigger the track visibility change.
- For local track we rely on 2 input signals to disable the track a) we disable locally the track we have a reference to, b) we send a mute request to the SFU which replies with an event that updates the participant to `hasVideo:false`. Due to this we have multiple instances of VideoRendererView trying to use the cached VideoRenderer. 

### 🛠 Implementation

- We now remove the unpublished track only in the case where the reason is different that `muted`
- We apply the visibility modifier on all renderer views regardless of the user's publishing video.
- We now have an optimisation in the VideoRendererView, where if the participant doesn't need to `showVideo` we render a dummy VideoRenderer and not trying to use the cached value.

### 🧪 Manual Testing Notes

- Enter a call from web and iOS (with more than 6 participants).
- Try all layouts and verify they work.
- Start sharing from web and ensure that the local track is visible in all layouts.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideoSwiftUI/CallView/VisibilityThresholdModifier.swift
+++ b/Sources/StreamVideoSwiftUI/CallView/VisibilityThresholdModifier.swift
@@ -11,7 +11,7 @@ import SwiftUI
 /// "on screen".
 ///
 /// When the visibility state of the view changes (i.e., it transitions between being "on screen" and "off screen"),
-/// a callback is triggered to notify the user of this change. This can be particularly useful in scenarios where 
+/// a callback is triggered to notify the user of this change. This can be particularly useful in scenarios where
 /// resource management is crucial, such as video playback or dynamic content loading, where actions might
 /// be triggered based on whether a view is currently visible to the user.
 ///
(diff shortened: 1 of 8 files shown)


martinmitrevski (approved):

Sources/StreamVideoSwiftUI/CallView/VisibilityThresholdModifier.swift
```diff
@@ -72,10 +72,10 @@ struct VisibilityThresholdModifier: ViewModifier {
 
         /// Check if the content view is vertically within the parent's bounds.
         let verticalVisible = (minY + requiredHeight <= bounds.maxY && minY >= bounds.minY) ||
-                              (maxY - requiredHeight >= bounds.minY && maxY <= bounds.maxY)
+        (maxY - requiredHeight >= bounds.minY && maxY <= bounds.maxY)
```
martinmitrevski, line 75: Indentation here and on the next one

martinmitrevski (approved):
Nice find 👍  So now we remove the track in 2 cases:
- when participant leaves the call
- when the track is unpublished for other reasons than muting

Sources/StreamVideoSwiftUI/CallView/VisibilityThresholdModifier.swift
ipavlidakis: I actually did that in purpose to help with the readability of the calculations. I can roll it back if you want though.

PR #283, opened 2024-02-02
[Feature]Capture CallView snapshot

ipavlidakis:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/714

### 🎯 Goal

Provide a way to capture snapshots of the CallView and pass it through the CallViewModel's flow.

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/Sources/Components/SnapshotTrigger.swift
+++ b/DemoApp/Sources/Components/SnapshotTrigger.swift
@@ -2,18 +2,36 @@
 // Copyright © 2024 Stream.io Inc. All rights reserved.
 //
 
+import Combine
 import Foundation
 import StreamVideo
+import StreamVideoSwiftUI
+import SwiftUI
 
- final class StreamSnapshotTrigger: ObservableObject {
+final class StreamSnapshotTrigger: SnapshotTriggering {
+    lazy var binding: Binding<Bool> = Binding<Bool>(
+        get: { [weak self] in
+            self?.currentValueSubject.value ?? false
+        },
+        set: { [weak self] in
+            self?.currentValueSubject.send($0)
+        }
+    )
 
-    @Published var capture: Bool = false
+    var publisher: AnyPublisher<Bool, Never> { currentValueSubject.eraseToAnyPublisher() }
+
+    private let currentValueSubject = CurrentValueSubject<Bool, Never>(false)
 
     init() {}
+
+    func capture() {
+        binding.wrappedValue = true
+    }
 }
 
 /// Provides the default value of the `StreamSnapshotTrigger` class.
 struct StreamSnapshotTriggerKey: InjectionKey {
+    @MainActor
     static var currentValue: StreamSnapshotTrigger = .init()
 }
 
@@ -28,4 +46,3 @@ extension InjectedValues {
         }
     }
 }
-
--- a/DemoApp/Sources/Components/SnapshotTrigger.swift
+++ b/DemoApp/Sources/Components/SnapshotTrigger.swift
@@ -1,8 +1,5 @@
 //
-//  SnapshotTrigger.swift
-//  StreamVideoSwiftUI
-//
-//  Created by Ilias Pavlidakis on 2/2/24.
+// Copyright © 2024 Stream.io Inc. All rights reserved.
 //
 
 import Foundation
--- a/DemoApp/Sources/Components/SnapshotTrigger.swift
+++ b/DemoApp/Sources/Components/SnapshotTrigger.swift
@@ -1,0 +1,34 @@
+//
+//  SnapshotTrigger.swift
+//  StreamVideoSwiftUI
+//
+//  Created by Ilias Pavlidakis on 2/2/24.
+//
+
+import Foundation
+import StreamVideo
+
+ final class StreamSnapshotTrigger: ObservableObject {
+
+    @Published var capture: Bool = false
+
+    init() {}
+}
+
+/// Provides the default value of the `StreamSnapshotTrigger` class.
+struct StreamSnapshotTriggerKey: InjectionKey {
+    static var currentValue: StreamSnapshotTrigger = .init()
+}
+
+extension InjectedValues {
+    /// Provides access to the `StreamSnapshotTrigger` class to the views and view models.
+    var snapshotTrigger: StreamSnapshotTrigger {
+        get {
+            Self[StreamSnapshotTriggerKey.self]
+        }
+        set {
+            Self[StreamSnapshotTriggerKey.self] = newValue
+        }
+    }
+}
+
--- a/DemoApp/Sources/Views/CallView/DemoCallView.swift
+++ b/DemoApp/Sources/Views/CallView/DemoCallView.swift
@@ -11,6 +11,7 @@ struct DemoCallView<ViewFactory: DemoAppViewFactory>: View {
 
     @Injected(\.appearance) var appearance
     @Injected(\.chatViewModel) var chatViewModel
+    @Injected(\.snapshotTrigger) var snapshotTrigger
 
     var microphoneChecker: MicrophoneChecker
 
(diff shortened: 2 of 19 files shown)


martinmitrevski (commented):
Not sure about the approach here: it implies that customers need to use our UI components + call view model to have this available.
Another possibility - have e.g. `call.snapshot(trackId:) async -> UIImage` that will take the current pixel buffer and convert it to an image - for the provided track id. Wdyt?

DemoApp/Sources/Views/CallView/DemoCallView.swift
```diff
@@ -76,6 +77,20 @@ struct DemoCallView<ViewFactory: DemoAppViewFactory>: View {
                 updateMicrophoneChecker()
             }
             .chat(viewModel: viewModel, chatViewModel: chatViewModel)
+            .snapshot(captureTrigger: snapshotTrigger.$capture.eraseToAnyPublisher()) { [weak viewModel, weak snapshotTrigger] in
+                guard let data = $0.pngData() else { return }
```
martinmitrevski, line 81: maybe we could use jpeg with 0.3 compression and add a comment about the size limitation.

DemoApp/Sources/Components/SnapshotTrigger.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  SnapshotTrigger.swift
```
martinmitrevski, line 2: Stream header.

DemoApp/Sources/Views/CallView/DemoCallView.swift
```diff
@@ -76,6 +77,20 @@ struct DemoCallView<ViewFactory: DemoAppViewFactory>: View {
...
+                snapshotTrigger?.capture = false
+                Task {
+                    do {
+                        let response = try await viewModel?.call?.sendCustomEvent([
+                            "snapshot": .string(data.base64EncodedString())
```
martinmitrevski, line 86: I think customers can even send the string split up, and then they reconstruct it on the receiving side (e.g. if offset and limit are provided).

DemoApp/Sources/Views/Controls/DemoControls.swift
```diff
@@ -164,3 +165,31 @@ extension View {
...
+    let size: CGFloat
+
+    init(size: CGFloat = 44) {
+        self.size = size
+    }
+
+    var body: some View {
+        Button(
+            action: {
+                snapshotTrigger.capture = true
```
martinmitrevski, line 183: I think this is a bit error prone atm - you can forget to set it to `false` for example. Ideally, this would be a method `capture()`, which will handle all the manipulation of the state. Although it's a demo app implementation, I think it's good if we improve this, wdyt?

martinmitrevski (commented):
Looks good, I think we can go into this direction 👍  Added few comments, plus we also need to update the docs.

DemoApp/Sources/Views/CallView/DemoCallView.swift
ipavlidakis: We discussed that offline. That will be complicated to implement as splitting the string into multiple it requires handling in order to stitch parts correctly in order, handle delays and even missed events. With the infrastructure in place and some mentioned restrictions, it will be on customers side to implement that.

DemoApp/Sources/Views/CallView/DemoCallView.swift
ipavlidakis: I'm not sure i agree. Even though what you are saying may be valid for a specific usecase it may fail for others (as it will return a low-resolution image). IMO it makes sense return something that has full data and customers can lower the resolution as they need during their processing.

DemoApp/Sources/Components/SnapshotTrigger.swift
ipavlidakis: Surprising why swift-format didn't catch it 🤔

DemoApp/Sources/Views/Controls/DemoControls.swift
ipavlidakis: The reason this is like that is because the control and actually viewHierarchy we need to snapshot are in different places. Without making changes in our UI infra i couldn't figure out a way to do it. Do you have any suggestions?

DemoApp/Sources/Components/SnapshotTrigger.swift
ipavlidakis: Found it #285

DemoApp/Sources/Views/CallView/DemoCallView.swift
martinmitrevski: The concern I have is that this request will almost never work - snapshot images would usually be over the limit. Another option I can think of is to actually not display it as part of the demo app.

DemoApp/Sources/Views/CallView/DemoCallView.swift
martinmitrevski: Yes, that makes sense. Maybe we can write few sentences in the custom event docs explaining this?

DemoApp/Sources/Views/Controls/DemoControls.swift
martinmitrevski: Yes, it's a bit tricky with the current setup. One idea could be: maybe we change `captureTrigger: AnyPublisher<Bool, Never>` to be a `@Binding`, since it's used in a SwiftUI view? That way when the snapshot is generated, we just reset it to `false` - similar to how `NavigationLink` works.

DemoApp/Sources/Views/CallView/DemoCallView.swift
ipavlidakis: I think you have in your mind only the scenario where someone wants to send the snapshot via a custom-event. There are other scenarios like sharing it to chat or other applications where reducing quality will affect them. 

> Another option I can think of is to actually not display it as part of the demo app.
I'm not sure i understand this part. What option are you referring to? Currently we aren't displaying this image anywhere. We only did this part of the as an example.

Sources/StreamVideoSwiftUI/CallViewModel.swift
```diff
@@ -445,7 +445,11 @@ open class CallViewModel: ObservableObject {
     public func update(participantsSortComparators: [StreamSortComparator<CallParticipant>]) {
         self.participantsSortComparators = participantsSortComparators
     }
-    
+
+    public func didCaptureSnapshot(_ snapshot: UIImage) {
```
martinmitrevski, line 449: Btw, is this used?

DemoApp/Sources/Views/CallView/DemoCallView.swift
ipavlidakis: It sounds too random or too specific to add it in the custom events. WDYT about a separate guide? But in order to do it we need probably a more complete example (we will need to provide at least something to do with this image).

DemoApp/Sources/Views/Controls/DemoControls.swift
ipavlidakis: But then someone can forget to set it to false and that would bring us to the same situation, right? The difference with the publisher is that it's a stream and doesn't affect rendering. While changing the value of a binding will cause re-renders.

DemoApp/Sources/Views/CallView/DemoCallView.swift
martinmitrevski: Well the sample code here sends custom events, and as far as I can see, the button to trigger this flow is visible. So what will happen is that whenever someone presses that button, the "limit" error will be thrown, right? If yes, then what's the point of having it visible or not having a size reduction?

DemoApp/Sources/Views/CallView/DemoCallView.swift
martinmitrevski: Maybe just mention that there's a limit in the custom event section? I don't think we have that documented. Then we can say if you want to send larger stuff, you can either upload it somewhere or send it in chunks.

DemoApp/Sources/Views/Controls/DemoControls.swift
martinmitrevski: If it's binding, we can change it directly in `SnapshotViewContainer`, when the snapshot is generated, right? Re-renders are not a problem here, this value will change only on user action.

DemoApp/Sources/Views/CallView/DemoCallView.swift
ipavlidakis: I just got what you were saying. This isn't a change on the SDK but on the DemoApp (the caller). It definitely makes sense to reduce it.

DemoApp/Sources/Views/Controls/DemoControls.swift
ipavlidakis: The problem with the re-renders is that we end up capturing an empty view (even with a small delay).

martinmitrevski (approved):

docusaurus/docs/iOS/05-ui-cookbook/16-snapshot.mdx
```diff
@@ -0,0 +1,209 @@
+---
+title: Snapshot
```
martinmitrevski, line 2: Just Snapshot is a bit generic, maybe "Video Feed Snapshot" or "Capturing Snapshots"?

martinmitrevski (approved):
Looks great 👌  Few small suggestions in the docs, but no need to wait the CI again.

PR #284, opened 2024-02-02
[Enhancement]Improve device orientation changes

ipavlidakis:
### 🎯 Goal

Provide a unified and platform agnostic way to determine device orientation.

### 🛠 Implementation

- DeviceOrientation across the SDK wasn't consistent. In a few places we were using `UIDeviceOrientation` while in others `UIInterfaceOrientation`. As those are platform specific and contain more information than the SDK needs, this PR attempts to replace the platform-specific values with an agnostic one and provide via the internal Dependency Injection infrastructure, an object that can be observed for orientation changes. Additionally it allows for customers to replace the object with their own implementation that fits their use case better, if needs be.

- This PR removes the `onRotate` ViewModifier and replaces it with the `orientationAdapter` Injected value.

### 🧪 Manual Testing Notes

- Enter a call with different number of users and rotate the device many times.

### ☑️ Contributor Checklist


martinmitrevski (commented):
Very nice improvement 👍 Please fix the tests before merging, since they still rely on the `orientation` parameter.

PR #290, opened 2024-02-07
Fix small snapshot issues

ipavlidakis:
### 📝 Summary

Fix some small issues when using the snapshot modifier and updating the relevant docs.

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/Sources/Extensions/CallViewModel+Snapshot.swift
+++ b/DemoApp/Sources/Extensions/CallViewModel+Snapshot.swift
@@ -5,10 +5,18 @@
 import Foundation
 import StreamVideo
 import StreamVideoSwiftUI
+import UIKit
 
 extension CallViewModel {
 
-    func sendSnapshot(_ snapshotData: Data) {
+    func sendSnapshot(_ snapshot: UIImage) {
+        guard
+            let resizedImage = resize(image: snapshot, to: .init(width: 30, height: 30)),
+            let snapshotData = resizedImage.jpegData(compressionQuality: 0.8)
+        else {
+            return
+        }
+
         Task {
             do {
                 try await call?.sendCustomEvent([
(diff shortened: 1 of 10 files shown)


DemoApp/Sources/Extensions/CallViewModel+Snapshot.swift
```diff
@@ -20,6 +28,43 @@ extension CallViewModel {
             }
         }
     }
+
+    private func resize(
```
laevandus, line 32: Would it make sense to have it inside a separate `ImageScaler` type?

DemoApp/Sources/Extensions/CallViewModel+Snapshot.swift
```diff
@@ -20,6 +28,43 @@ extension CallViewModel {
...
+            width: scaledWidth,
+            height: scaledHeight
+        )
+
+        // Create a new image context
+        UIGraphicsBeginImageContextWithOptions(targetSize, false, 0)
+        image.draw(in: targetRect)
+
+        let newImage = UIGraphicsGetImageFromCurrentImageContext()
+        UIGraphicsEndImageContext()
```
laevandus, line 64: Since we turn this into a JPEG in the next step then I was wondering if we could benefit from: [UIGraphicsImageRenderer.jpegData(withCompressionQuality:actions:)](https://developer.apple.com/documentation/uikit/uigraphicsimagerenderer/1649234-jpegdata)
But in that case we probably should have the targetRect logic separated as well for making it easier to use it with jpeg creation method and the image creation method on UIGraphicsImageRenderer.

DemoApp/Sources/Extensions/CallViewModel+Snapshot.swift
```diff
@@ -20,6 +28,43 @@ extension CallViewModel {
...
+
+        // Determine the scale factor that preserves aspect ratio
+        let scaleFactor = min(widthRatio, heightRatio)
+
+        let scaledWidth = image.size.width * scaleFactor
+        let scaledHeight = image.size.height * scaleFactor
+        let targetRect = CGRect(
+            x: (
+                targetSize.width - scaledWidth
+            ) / 2,
```
laevandus, line 53: Formatting looks a little bit odd here compared to `y`

DemoApp/Sources/Extensions/CallViewModel+Snapshot.swift
ipavlidakis: The reasoning for keeping here is in order to make it easier to write the documentation and have something to point for integrators.

DemoApp/Sources/Extensions/CallViewModel+Snapshot.swift
ipavlidakis: Later on we are using it (via UIImage) but we don't have "hard" requirements here. We are mostly doing it demonstrate sending data through our WebRTC stack (which poses a 100Kb size limitation and thus we the reduction in quality).

DemoApp/Sources/Extensions/CallViewModel+Snapshot.swift
ipavlidakis: I'm wondering why swift-format didn't catch it. That happened because i pressed ctrl+M on the line. I'll update it

martinmitrevski (approved):

laevandus (approved):

PR #303, opened 2024-02-19
[Fix]Single instance AVAudioRecorder

ipavlidakis:
### 🎯 Goal

AVAudioRecorder is an expensive resource. This PR aims to centralise its usage and maintain a single instance during runtime, that will be initialized once and early to avoid delays during the call.

### 📝 Summary

- Abstract the AVAudioRecorder behind a specialized, thread safe object.
- Inject it into the UI components and update paths accordingly.
- Minor UI updates to make the interface smoother.
- Extensive unit testing for the recording layer.

### 🧪 Manual Testing Notes

|  Lobby  |  Join Call  | Leave Call|
| ------- | ----------- | --------- |
|  true   |  true       | false     |
|  false  |  false      | false     |
|  false  |  true       | true      |
|  true   |  false      | false     |

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/Sources/Views/CallView/DemoCallView.swift
+++ b/DemoApp/Sources/Views/CallView/DemoCallView.swift
@@ -38,7 +38,7 @@ struct DemoCallView<ViewFactory: DemoAppViewFactory>: View {
         viewFactory
             .makeInnerCallView(viewModel: viewModel)
             .onReceive(viewModel.callSettingsPublisher) { _ in
-                updateMicrophoneChecker()
+                Task { await updateMicrophoneChecker() }
             }
             .onReceive(microphoneChecker.decibelsPublisher, perform: { values in
                 guard !viewModel.callSettings.audioOn else { return }
@@ -72,22 +72,16 @@ struct DemoCallView<ViewFactory: DemoAppViewFactory>: View {
                         : nil
                 }
             )
-            .onDisappear {
-                microphoneChecker.stopListening()
-            }
-            .onAppear {
-                updateMicrophoneChecker()
-            }
             .presentsMoreControls(viewModel: viewModel)
             .chat(viewModel: viewModel, chatViewModel: chatViewModel)
             .toastView(toast: $snapshotViewModel.toast)
     }
 
-    private func updateMicrophoneChecker() {
-        if viewModel.call != nil, !viewModel.callSettings.audioOn {
-            microphoneChecker.startListening()
-        } else {
-            microphoneChecker.stopListening()
+    private func updateMicrophoneChecker() async {
+        if viewModel.call != nil, viewModel.callSettings.audioOn {
+            await microphoneChecker.startListening()
+        } else if viewModel.call != nil, !viewModel.callSettings.audioOn {
+            await microphoneChecker.stopListening()
         }
     }
 }
--- a/Sources/StreamVideo/Utils/AudioRecorder/AVAudioRecorderBuilder.swift
+++ b/Sources/StreamVideo/Utils/AudioRecorder/AVAudioRecorderBuilder.swift
@@ -5,6 +5,12 @@
 import AVFoundation
 import Foundation
 
+/// The AVAudioRecorderBuilder actor simplifies the creation and management of AVAudioRecorder
+/// instances for audio recording. It offers:
+/// Caching: Stores created AVAudioRecorder objects for efficient reuse, avoiding redundant initialization.
+/// Customisable settings: Allows you to tailor recording parameters to your specific needs.
+///
+/// - Important: You need to call `.build()` before trying to access the `result` property.
 actor AVAudioRecorderBuilder {
 
     // `kAudioFormatLinearPCM` is being used to be able to support multiple
@@ -18,13 +24,14 @@ actor AVAudioRecorderBuilder {
         AVEncoderAudioQualityKey: AVAudioQuality.high.rawValue
     ]
 
+    /// The URL where the audio recording will be saved.
     let fileURL: URL
+
+    /// A dictionary containing audio format, sample rate, number of channels, and quality configurations.
     let settings: [String: Any]
 
+    /// A property storing the built AVAudioRecorder instance.
     private var cachedResult: AVAudioRecorder?
-    private var factory: (URL, [String: Any]) throws -> AVAudioRecorder = {
-        try .init(url: $0, settings: $1)
-    }
 
     var result: AVAudioRecorder? { cachedResult }
 
@@ -48,9 +55,10 @@ actor AVAudioRecorderBuilder {
         self.settings = cachedResult.settings
     }
 
+    /// Instructs the `AVAudioRecorderBuilder` to build and cache an instance of AVAudioRecorder.
     func build() throws {
         guard cachedResult == nil else { return }
-        let audioRecorder = try factory(fileURL, settings)
+        let audioRecorder = try AVAudioRecorder(url: fileURL, settings: settings)
         self.cachedResult = audioRecorder
     }
 }
--- a/Sources/StreamVideo/Utils/AudioRecorder/AVAudioRecorderBuilder.swift
+++ b/Sources/StreamVideo/Utils/AudioRecorder/AVAudioRecorderBuilder.swift
@@ -1,0 +1,58 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import AVFoundation
+import Foundation
+
+actor AVAudioRecorderBuilder {
+
+    // `kAudioFormatLinearPCM` is being used to be able to support multiple
+    // instances of AVAudioRecorders. (useful when using MicrophoneChecker
+    // during a Call).
+    // https://stackoverflow.com/a/8575101
+    static let defaultRecordingSettings: [String: any Sendable] = [
+        AVFormatIDKey: Int(kAudioFormatLinearPCM),
+        AVSampleRateKey: 12000,
+        AVNumberOfChannelsKey: 1,
+        AVEncoderAudioQualityKey: AVAudioQuality.high.rawValue
+    ]
+
+    let fileURL: URL
+    let settings: [String: Any]
+
+    private var cachedResult: AVAudioRecorder?
+    private var factory: (URL, [String: Any]) throws -> AVAudioRecorder = {
+        try .init(url: $0, settings: $1)
+    }
+
+    var result: AVAudioRecorder? { cachedResult }
+
+    init(
+        inCacheDirectoryWithFilename filename: String,
+        settings: [String: any Sendable] = AVAudioRecorderBuilder.defaultRecordingSettings
+    ) {
+        let documentPath = FileManager.default.urls(
+            for: .cachesDirectory,
+            in: .userDomainMask
+        )[0]
+        self.fileURL = documentPath.appendingPathComponent(filename)
+        self.settings = settings
+    }
+
+    init(
+        cachedResult: AVAudioRecorder
+    ) {
+        self.cachedResult = cachedResult
+        self.fileURL = cachedResult.url
+        self.settings = cachedResult.settings
+    }
+
+    func build() throws {
+        guard cachedResult == nil else { return }
+        let audioRecorder = try factory(fileURL, settings)
+        self.cachedResult = audioRecorder
+    }
+}
+
+extension AVAudioRecorder: @unchecked Sendable {}
--- a/Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
+++ b/Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
@@ -1,0 +1,180 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import AVFoundation
+import Combine
+import Foundation
+
+/// This class abstracts the usage of AVAudioRecorder, providing a convenient way to record and manage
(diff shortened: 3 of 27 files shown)


Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
```diff
@@ -0,0 +1,180 @@
...
+            } catch {
+                if type(of: error) != CancellationError.self {
+                    log.error("Failed to create AVAudioRecorder.", error: error)
+                }
+            }
+        }
+    }
+
+    private func setUpAudioCaptureIfRequired(
+        _ completionHandler: @escaping (AVAudioRecorder) async -> Void
```
martinmitrevski, line 147: why do we need completion handler here? Can't we return the recorder directly after the async stuff finishes?

Sources/StreamVideo/Utils/AudioRecorder/AVAudioRecorderBuilder.swift
```diff
@@ -0,0 +1,58 @@
...
+actor AVAudioRecorderBuilder {
+
+    // `kAudioFormatLinearPCM` is being used to be able to support multiple
+    // instances of AVAudioRecorders. (useful when using MicrophoneChecker
+    // during a Call).
+    // https://stackoverflow.com/a/8575101
+    static let defaultRecordingSettings: [String: any Sendable] = [
+        AVFormatIDKey: Int(kAudioFormatLinearPCM),
+        AVSampleRateKey: 12000,
+        AVNumberOfChannelsKey: 1,
```
laevandus, line 17: Low bitrate and mono audio is currently working as designed?

Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
```diff
@@ -0,0 +1,180 @@
...
+    /// A public publisher that exposes the average power of the audio signal.
+    open private(set) lazy var metersPublisher: AnyPublisher<Float, Never> = _metersPublisher.eraseToAnyPublisher()
+
+    /// Indicates whether an active call is present, influencing recording behavior.
+    public var hasActiveCall: Bool = false {
+        didSet {
+            if !hasActiveCall {
+                Task {
+                    await stopRecording()
+                    try? audioSession.setActive(false, options: [])
```
laevandus, line 39: Should we log an error here?

DemoApp/Sources/Views/CallView/DemoCallView.swift
```diff
@@ -38,7 +38,7 @@ struct DemoCallView<ViewFactory: DemoAppViewFactory>: View {
         viewFactory
             .makeInnerCallView(viewModel: viewModel)
             .onReceive(viewModel.callSettingsPublisher) { _ in
-                updateMicrophoneChecker()
+                Task { await updateMicrophoneChecker() }
```
laevandus, line 41: Not important for now but maybe at some point we should add a async supported `onReceive`.

Sources/StreamVideoSwiftUI/CallViewModel.swift
```diff
@@ -71,6 +72,7 @@ open class CallViewModel: ObservableObject {
     @Published public var callingState: CallingState = .idle {
         didSet {
             handleRingingEvents()
+            audioRecorder.hasActiveCall = callingState != .idle
```
martinmitrevski, line 75: this part can be tricky a bit for customers that won't use our view model, can we manage it internally instead in some way?

Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
```diff
@@ -0,0 +1,180 @@
...
+                if type(of: error) != CancellationError.self {
+                    log.error("Failed to create AVAudioRecorder.", error: error)
+                }
+            }
+        }
+    }
+
+    private func setUpAudioCaptureIfRequired(
+        _ completionHandler: @escaping (AVAudioRecorder) async -> Void
+    ) async {
```
laevandus, line 148: Slightly hard to read since it mixes `completionHandler` with `async`. Wondering if we could do async getter for returning the cached audio recorder.

Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
```diff
@@ -0,0 +1,180 @@
...
+                        self._metersPublisher.send(audioRecorder.averagePower(forChannel: 0))
+                        log.debug("️Recording meters updated")
+                    }
+                }
+        }
+    }
+
+    /// Stops recording audio asynchronously.
+    open func stopRecording() async {
+        guard let audioRecorder = await audioRecorderBuilder.result, audioRecorder.isRecording else {
```
martinmitrevski, line 109: thinking if we should do some cleanup of the file also when we stop recording (not only on deinit), wdyt?

Sources/StreamVideoSwiftUI/Utils/Camera.swift
```diff
@@ -129,38 +129,28 @@ class Camera: NSObject, @unchecked Sendable {
...
-        let photoOutput = AVCapturePhotoOutput()
-                        
-        captureSession.sessionPreset = AVCaptureSession.Preset.photo
+
+        captureSession.sessionPreset = AVCaptureSession.Preset.high
 
         let videoOutput = AVCaptureVideoDataOutput()
-        videoOutput.setSampleBufferDelegate(self, queue: DispatchQueue(label: "VideoDataOutputQueue"))
-  
+        videoOutput.setSampleBufferDelegate(self, queue: DispatchQueue.global(qos: .userInteractive))
```
laevandus, line 136: `setSampleBufferDelegate` says that "You must pass a serial dispatch to guarantee that audio samples will be delivered in order."

Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
```diff
@@ -0,0 +1,180 @@
...
+
+import AVFoundation
+import Combine
+import Foundation
+
+/// This class abstracts the usage of AVAudioRecorder, providing a convenient way to record and manage
+/// audio streams. It handles setting up the recording environment, starting and stopping recording, and
+/// publishing the average power of the audio signal. Additionally, it adjusts its behavior based on the
+/// presence of an active call, automatically stopping recording if needed.
+open class StreamAudioRecorder: @unchecked Sendable {
```
martinmitrevski, line 13: Do you think this might collide with the voice recording feature from chat?

martinmitrevski (commented):
Great work 👌  Let's test a bit for the upcoming meetings. Added few comments, but nothing major.

Sources/StreamVideo/Utils/AudioRecorder/AVAudioRecorderBuilder.swift
ipavlidakis: Yeap :)

Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
ipavlidakis: TBH it doesn't make much difference because either way our call has ended and this is only to mark the system's session (which any way will be deactivated after some time idle).

DemoApp/Sources/Views/CallView/DemoCallView.swift
ipavlidakis: That's a nice idea

Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
ipavlidakis: Very good point. Thanks! Updated

Sources/StreamVideoSwiftUI/Utils/Camera.swift
ipavlidakis: Very good poing! :lifesaver:

Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
ipavlidakis: It won't hurt for sure

Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
ipavlidakis: It does :/ Given that they are in different frameworks they will still be accessed correctly. Just for convenience would you prefer renaming it to `StreamCallAudioRecorder`?

Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
ipavlidakis: In any way it won't hurt for sure :) Let me add it

Sources/StreamVideoSwiftUI/CallViewModel.swift
martinmitrevski: This should be internal - whenever we enter a call we should update it in the SDK (shouldn't be available outside the LLC). Maybe we can manage it via the call controller?

Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
martinmitrevski: I'm not worried about the naming - the namespaces will handle that. More about the usage - e.g. if you send a voice message on the call, or even before/after.

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: I updated it. WDYT?

Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
ipavlidakis: It shouldn't have any different effect than it does now. As it seems we don't have async messages on SwiftUI chat and i can't test it. I believe that a custom handling for the AudioSession will be required in order when an async message gets sent we don't disable the AudioSession.

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: Actually, if we handle it internally we have an issue with the LobbyView, because the call isn't active yet.

Sources/StreamVideoSwiftUI/Utils/Camera.swift
```diff
@@ -15,7 +15,11 @@ class Camera: NSObject, @unchecked Sendable {
     private var photoOutput: AVCapturePhotoOutput?
     private var videoOutput: AVCaptureVideoDataOutput?
     private var sessionQueue: DispatchQueue!
-    
+    private lazy var frameProcessingQueue: DispatchQueue = DispatchQueue(
+        label: "io.getstream.\(String(describing: type(of: self))).videoDataOutputQueue",
+        target: .global(qos: .userInteractive)
+    )
```
laevandus, line 21: Since the target queue is concurrent, then it feels like it still has timing issues 
Example
```
queue = DispatchQueue(label: "a", target: .global(qos: .userInteractive))
DispatchQueue.concurrentPerform(iterations: 10) { index in
  queue.async {
    print(index)
  }
}
```
prints something like
```
1
0
2
3
4
6
5
7
8
9
```
Shouldn't we just go for serial queue without a target?

ipavlidakis: I think the problem is with your code as you are doing concurrentPerform as this isn't how the delegate will be called. The approach that i'm using is actually the one Apple recommends in their docs
```
For serial tasks, set the target of your serial queue to one of the global concurrent queues. That way, you can maintain the serialized behavior of the queue while minimizing the number of separate queues creating threads.
```

laevandus (approved):
Looks good from my side

Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
martinmitrevski: We have async messages on SwiftUI - they are probably not turned on (`utils.composerConfig.isVoiceRecordingEnabled`). Might be good to test it.

Sources/StreamVideoSwiftUI/CallViewModel.swift
martinmitrevski: we can expose a public interface specifically for the lobby, while we still handle it internally. If you build your own lobby, you would have to manage it manually, which is not that bad.

ipavlidakis: What i did was to have a parameter in the startListening that you can pass to ignore the activeCall existence.

Sources/StreamVideo/Utils/AudioRecorder/StreamAudioRecorder.swift
ipavlidakis: Just tested it. In order to enable async messages on ChatSwiftUI we need to provide a custom AudioSessionConfiguration (like we do on UIKit). That requires an update on ChatSwiftUI. Until then we cannot async messages on the demo app.

martinmitrevski (approved):

PR #309, opened 2024-02-23
[Feature]Implement image and blur background filters

ipavlidakis:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/736

### 🎯 Goal

Implement basic background filtering.

### 📝 Summary

This PR provides simple background filter for blur and custom image.

### 🛠 Implementation

Leveraging the CoreML power we are using a person segmentation model, to extract the person from their background and apply the desired transformation.

### 🎨 Showcase

|  Blur  |  Image  |
| -------- | ------- |
|  IMG_2192 | IMG_2191 |

### 🧪 Manual Testing Notes

- Open the DemoApp
- Join a call
- Try the different available filters

### ☑️ Contributor Checklist


DemoApp/Sources/Components/AppState.swift
```diff
@@ -52,6 +56,15 @@ final class AppState: ObservableObject {
...
             deferSetDevice = false
             deferSetVoipDevice = false
+            activeCallCancellable?.cancel()
+            activeCallCancellable = nil
+            if let streamVideo {
+                activeCallCancellable = streamVideo
+                    .state
+                    .$activeCall
+                    .receive(on: DispatchQueue.main)
+                    .sink { [weak self] in self?.activeCall = $0 }
```
laevandus, line 66: `assign(to:)` instead of `sink` could be nicer to read but in that case we would need something like `assignWeakly(to)`

```swift
extension Publisher where Self.Failure == Never {
public func assignNoRetain<Root>(to keyPath: ReferenceWritableKeyPath<Root, Self.Output>, on object: Root) -> AnyCancellable where Root: AnyObject {
    sink { [weak object] (value) in
        object?[keyPath: keyPath] = value
    }
  }
}
```

DemoApp/Sources/Controls/DemoMoreControls/DemoBackgroundEffectSelector.swift
```diff
@@ -0,0 +1,119 @@
...
+
+import Foundation
+import StreamVideo
+import StreamVideoSwiftUI
+import SwiftUI
+
+@available(iOS 15.0, *)
+struct DemoBackgroundEffectSelector: View {
+
+    var effects: [BackgroundEffect] = BackgroundEffect.allCases
```
laevandus, line 13: Minor: `var` to `let` since we really can't mutate it like this?

(some other uses below as well)

DemoApp/Sources/Components/AppState.swift
ipavlidakis: yeah assign isn't great because it holds a strong reference. Assign weak is a common publisher extension but we don't have it. I think it's alright to leave with it (especially as its in the DemoApp and we prefer to keep things as vanilla as possible)

DemoApp/Sources/Controls/DemoMoreControls/DemoBackgroundEffectSelector.swift
```diff
@@ -0,0 +1,119 @@
...
+    case blur
+    case amsterdam1 = "amsterdam-1"
+    case amsterdam2 = "amsterdam-2"
+    case boulder1 = "boulder-1"
+    case boulder2 = "boulder-2"
+    case gradient1 = "gradient-1"
+    case gradient2 = "gradient-2"
+    case gradient3 = "gradient-3"
+
+    var id: ObjectIdentifier { ObjectIdentifier(rawValue as NSString) }
```
laevandus, line 79: Interesting that there is `as NSString`. Not obviously clear why but I guess it has reasons.

laevandus (approved):

DemoApp/Sources/Controls/DemoMoreControls/DemoBackgroundEffectSelector.swift
ipavlidakis: ObjectIdentifier requires an reference type and bridging a String to NSString is essentially a free operation I'm using it here to satisfy the compiler :)

Sources/StreamVideo/WebRTC/VideoFilters/Extensions/UIDevice+NeuralEngine.swift
```diff
@@ -0,0 +1,16 @@
...
+import MLCompute
+var neuralEngineExists = {
+    if #available(iOS 15.0, *) {
+        return MLCDevice.ane() != nil
+    } else {
+        return false
+    }
+}()
+#else
+var neuralEngineExists = false
```
martinmitrevski, line 15: probably safer to make this a let.

martinmitrevski (commented):
Fantastic work 👏  Before merging, we need to update the docs with examples. Also, the changelog should be updated.

martinmitrevski (approved):

PR #310, opened 2024-02-27
Update XCTest fulfillment function

testableapple:


TestTools/TestData/XCTestCase.swift
```diff
@@ -0,0 +1,19 @@
+//
```
ipavlidakis, line 1: Can we rename the file to `XCTestCase+SafeFulfillment`?

ipavlidakis (approved):

PR #334, opened 2024-03-21
[Proposal]CallKit handling on LLC

ipavlidakis:
### 🎯 Goal

Make CallKit integration simpler.

### 📝 Summary

The PR introduces a stack of objects that take care of registering with CallKit and managing every aspect of a call.

### 🛠 Implementation

Main objects:
- `CallKitAdapter` (Root): acts as an intermediary between the application and CallKit services, facilitating registration for incoming calls and managing the CallKit service instance.
- `CallKitPushNotificationAdapter`: Handles push notifications for CallKit integration.
- `CallKitService`: manages interactions with the CallKit framework, facilitating VoIP calls in an application.

All objects are managed via InjectedValues and are only initialized once referenced (you don't need to reference all of them just the Root one).

### 🎨 Showcase

Every step in this [doc](https://getstream.io/video/docs/ios/advanced/callkit-integration/#ios-app-integration) are being replaced from the following calls
```swift
// Once StreamVideo gets created we need to call 
InjectedValues[\.callKitAdapter].streamVideo = streamVideo

// From our Root View
InjectedValues[\.callKitAdapter].registerForIncomingCalls()

// Optionally we can pass in our logo to be displayed in the VoIP push notification
InjectedValues[\.callKitAdapter].iconTemplateImageData = UIImage(named: "logo")?.pngData()

```

### 🧪 Manual Testing Notes

- Use a couple of devices
- Sign in with different users
- Call from one device the users on the other devices
- Perform every available action on every app state (foregroud, background, not running)

### ☑️ Contributor Checklist


martinmitrevski (commented):
Looks great 👌 Before merging, let's fix the crash in `main`. Also, we need to update the docs a bit and do more testing.

Sources/StreamVideo/CallKit/CallKitAdapter.swift
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        loggedInStateCancellable = streamVideo
+            .state
+            .$connection
+            .sink { [weak self] in
+                switch $0 {
+                case .connected:
+                    self?.registerForIncomingCalls()
+                case .disconnected:
```
martinmitrevski, line 58: Not sure about this: if you put the app in the background, we will disconnect. In that case, we shouldn't unregister for incoming calls. Wdyt?

Sources/StreamVideoUIKit/CallViewController.swift
```diff
@@ -14,9 +14,9 @@ open class CallViewController: UIViewController {
     
     private var cancellables = Set<AnyCancellable>()
     
-    public static func make(with viewModel: CallViewModel? = nil) -> CallViewController {
```
martinmitrevski, line 17: Why is it mandatory to provide the view model now?

StreamVideoTests/CallKit/CallKitAdapterTests.swift
```diff
@@ -0,0 +1,131 @@
...
+
+        // When
+        subject.streamVideo = streamVideo
+
+        // Then
+        XCTAssertTrue(callKitService.streamVideo === streamVideo)
+        XCTAssertTrue(callKitPushNotificationAdapter.registerWasCalled)
+    }
+
+    func testDidUpdate_streamVideoIsNotNilAndConnected_whenItDisconnect_callKitServiceWasUpdatedAndUnRegisterWasCalled(
```
martinmitrevski, line 74: nit: a bit too long name

StreamVideoTests/CallKit/CallKitServiceTests.swift
```diff
@@ -0,0 +1,197 @@
...
+        // Then
+        waitForExpectations(timeout: defaultTimeout, handler: nil)
+        XCTAssertNil(completionError)
+        XCTAssertTrue(callProvider.reportNewIncomingCallCalled)
+        XCTAssertEqual(callProvider.reportNewIncomingCallUpdate?.localizedCallerName, localizedCallerName)
+        XCTAssertEqual(callProvider.reportNewIncomingCallUpdate?.remoteHandle?.value, callerId)
+    }
+
+    func test_reportIncomingCall_streamVideoIsNil_callWasEnded() async throws {
+
```
martinmitrevski, line 58: Why do we have empty space before each of these tests? And why is not the linter complaining 😅

StreamVideoTests/Utilities/Dummy/CallResponse+Dummy.swift
```diff
@@ -0,0 +1,54 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamVideo
+
+extension CallResponse {
```
martinmitrevski, line 8: I think we already had a way to create some of these responses.

StreamVideoTests/Utilities/Dummy/ScreensharingSettings+Dummy.swift
```diff
@@ -0,0 +1,15 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamVideo
+
+extension ScreensharingSettings {
```
martinmitrevski, line 8: For some of these (like this one) feels like an overkill to have a separate file and a `dummy` method.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - The return type of `call.get()` is now the API type `GetCallResponse` which encapsulates the previous `CallResponse` under the `call` property. [#335](https://github.com/GetStream/stream-video-swift/pull/335)
 - Remove Nuke dependency from the SwiftUI SDK. [#340](https://github.com/GetStream/stream-video-swift/pull/340)
 
+### ✅ Added
+- The SDK now provides a mechanism to manage CallKit integration. [#334](https://github.com/GetStream/stream-video-swift/pull/334)
```
martinmitrevski, line 12: Can we re-word this to be more specific? As in - we provide helpers inside the SDK to help with your CallKit integration, sth around those lines.

martinmitrevski (commented):
Looks great 👍  Added few comments, let me know what you think. Also, we still need the updated docs.

StreamVideoTests/Utilities/Dummy/CallResponse+Dummy.swift
ipavlidakis: Really? Can you elaborate?

StreamVideoTests/Utilities/Dummy/ScreensharingSettings+Dummy.swift
ipavlidakis: It's easier to scan the project and see what has a mock and IMO they are clearer. I don't really understand the term overkill as those files are in tests under a specific folder named Mocks and they don't complicate something.

CHANGELOG.md
ipavlidakis: Maybe we should do better marketing here as we are not helping with the CallKit integration rather we are doing it completely. 

WDYT about "The SDK now provides a CallKit integration if needed."?

StreamVideoTests/Utilities/Dummy/CallResponse+Dummy.swift
martinmitrevski: check the `MockResponseBuilder`

StreamVideoTests/Utilities/Dummy/ScreensharingSettings+Dummy.swift
martinmitrevski: For me, it's an overkill to create new files and extensions just for passing one default bool parameter. For larger initializers it makes sense, but not for most of these. 
In any case, up to you.

Sources/StreamVideoUIKit/CallViewController.swift
ipavlidakis: I brought this one back to not break customers integration but i basically would like to deprecate it for a few reasons:
1. There is no reason to have optionality for the viewModel as internally we always ensure that there will be one
2. Static methods like this one aren't really swifty and are more difficult to be discovered (and require mention and explanation in docs). We can rely on convenience initializers to achieve the same result

StreamVideoTests/Utilities/Dummy/ScreensharingSettings+Dummy.swift
ipavlidakis: I think that maintaining seperate files per file and also keeping the `dummy` format we have on Chat is benefitial. While looking into the `MockResponseBuilder` i realised that the file is also shared between Video and VideoUI only by Xcode settings, which means that the Package itself will fail to run tests. 
As a final improvement i believe we should move those mocks in to a seperate TestHelpers target as we do for Chat.

martinmitrevski: Yes, good point, we should create a new target for these.

martinmitrevski (approved):
LGTM ✅

PR #350, opened 2024-04-11
[Improvement]Implement VideoRendererPool

ipavlidakis:
### 🎯 Goal

Eliminate tracks disappearing due to reusability on SwiftUI view renders.

### 🛠 Implementation

Adds an implementation for a ReusePool that can store components and reuse them based on needs. The pool can stretch and/or tight its capacity automatically.

### 🧪 Manual Testing Notes

- Get in a call with a few participants
- Try out all layouts
- Try out screensharing
- Try out PiP

### ☑️ Contributor Checklist


Sources/StreamVideoSwiftUI/CallView/VideoRenderer.swift
```diff
@@ -83,9 +84,7 @@ public struct VideoRendererView: UIViewRepresentable {
     }
 
     public func makeUIView(context: Context) -> VideoRenderer {
-        let view = showVideo
-            ? utils.videoRendererFactory.view(for: id, size: size)
-            : VideoRenderer()
+        let view = videoRendererPool.acquireRenderer(size: size)
```
martinmitrevski, line 87: Interesting approach - so the id is no longer taking into consideration. We just say this view is taken now, right?

Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
```diff
@@ -0,0 +1,59 @@
...
+            }
+    }
+
+    /// Acquires a `VideoRenderer` from the pool with the specified size.
+    ///
+    /// - Parameter size: The desired size for the acquired `VideoRenderer`.
+    /// - Returns: A `VideoRenderer` instance from the pool.
+    func acquireRenderer(size: CGSize) -> VideoRenderer {
+        let renderer = pool.acquire()
+        renderer.frame.size = size // Set the size of the renderer
```
martinmitrevski, line 38: Just to confirm - size is updated here and in the view lifecycle methods, right?

Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
```diff
@@ -0,0 +1,59 @@
...
+    /// A cancellable object to observe call end notifications for releasing all renderers.
+    private var callEndedCancellable: AnyCancellable?
+
+    /// Initializes the `VideoRendererPool` with a specified initial capacity.
+    ///
+    /// - Parameter initialCapacity: The initial capacity of the pool (default is 10).
+    init(initialCapacity: Int = 10) {
+        // Initialize the pool with a capacity and a factory closure to create `VideoRenderer` instances
+        pool = ReusePool(initialCapacity: initialCapacity) {
+            VideoRenderer(frame: CGRect(origin: .zero, size: .zero))
```
martinmitrevski, line 22: One downside is that we always create 10 views, even for direct calls.

Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        // Observe call end notifications to release all renderers when a call ends
+        callEndedCancellable = NotificationCenter.default.publisher(for: Notification.Name(CallNotification.callEnded))
+            .sink { [weak self] _ in
+                self?.pool.releaseAll()
```
martinmitrevski, line 28: If we had a call with 100 people, this will grow to 100 views, and they won't be removed for the next calls? (e.g. with only few people)

StreamVideoSwiftUITests/Utils/ReusePool_Tests.swift
```diff
@@ -0,0 +1,111 @@
+//
+//  ReusePool_Tests.swift
```
martinmitrevski, line 2: Stream header

martinmitrevski (commented):
Looks great 👍  I added few questions, let me know what you think. Also, let's test it on the big call today again.

Sources/StreamVideoSwiftUI/CallView/VideoRenderer.swift
ipavlidakis: Exactly. This is alright because our views are just observers on RTCVideoTrack instances that are built to deliver their frames to multiple observers. So we can get advantage of it simplify our lives by just saying, give a view i can render on and just make sure we free up resources when not in use(e.g. when views are being removed from the viewHierarchy)

Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
ipavlidakis: That's right, this basically replaces the functionality we had in the factory that was "Give me a videoRenderer of this size". There are no changes in the VideoRender layout code (which is the one responsible for updating the size on runtime)

Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
ipavlidakis: that's ok for 2 reasons: 1) VideoRenderers that don't have a track assigned on them are just plain views that do nothing 2) we can very easily make this configurable and even dynamically changed based on the active call (e.g. how many members are set in the Call that's the number of the videoRenderers we will create)(one more point here is that going lower won't hurt us as we the pool can expant to accomodate the call's needs)

Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
ipavlidakis: Actually the will be removed. `releaseAll` will maintain a number VideoRenders up to `initialCapacity` value and everything else will be released. In the scenario you mentioned, 10 VideoRenders will be marked as available and 90 will be released

StreamVideoSwiftUITests/Utils/ReusePool_Tests.swift
ipavlidakis: Oh that's interesting. I'm wondering with swift-format didn't catch it

Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
martinmitrevski: Cool, missed that part 👍

Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
martinmitrevski: No, I think it's fine like this. 10 UIViews won't impact the memory that much.

martinmitrevski (approved):
Amazing work 🚀

PR #392, opened 2024-05-15
[Enhancement]Allow CallKitService to handle multiple calls

ipavlidakis:
### 🎯 Goal

Simplify, verify and improve Calling state observation and reaction, from CallKit.

### 📝 Summary

The PR is improving the existing CallKitService by allowing it to handle multiple calls at the same time. This is quite important as while in a call, user may receive another incoming call. CallKitService should be able to reliably report each of those calls and differentiate user action on a call from the others.

### 🛠 Implementation

Calls are being received in a serial manner. For that purpose we are using a Stack implementation. The first entry on the stack will always be the latest call (active or ringing).

### 🎨 Showcase

https://github.com/GetStream/stream-video-swift/assets/472467/b6f06d61-ea0d-4549-b0a9-2848acc5c112

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/Sources/Components/AppEnvironment.swift
+++ b/DemoApp/Sources/Components/AppEnvironment.swift
@@ -144,7 +144,7 @@ extension AppEnvironment {
             return .detailed
         case .debug:
             #if targetEnvironment(simulator)
-            return .simple
+            return .detailed
             #else
             return .simple
             #endif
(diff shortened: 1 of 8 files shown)


martinmitrevski (approved):
Looks great ✅

PR #404, opened 2024-05-22
[Enhancement]Use single instances for Call objects with the same cId

ipavlidakis:
### 🎯 Goal

Use single instances for `Call` objects with the same `cId`. That helps ensuring that irregardless of the `Call` object creation point, actions performed to it will be synced across every observer.

### 📝 Summary

`StreamVideo` and `Call` internally use a thread-safe cache that ensures that only a single instance of `Call` object will exist for the same cId.

### 🧪 Manual Testing Notes

- Regular join call flows should remain working.

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### 🔄 Changed
+- `Call` objects for the same `cId` will reference the same memory instance. [#404](https://github.com/GetStream/stream-video-swift/pull/404)
+
 ### 🐞 Fixed
 - Video tracks for anonymous users not displaying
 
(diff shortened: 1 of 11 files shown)


Sources/StreamVideo/Utils/UnfairQueue/UnfairQueue.swift
```diff
@@ -23,15 +23,15 @@ import Foundation
 /// additional overhead of dispatching tasks and managing thread execution in a DispatchQueue could result
 /// in unnecessary latency, making `os_unfair_lock` the superior choice for scenarios where rapid, lightweight
 /// synchronization is paramount.
-final class UnfairQueue {
+public final class UnfairQueue {
```
martinmitrevski, line 26: Hmm, seeing this one again - it was also in the other PR.

martinmitrevski (approved):

Sources/StreamVideo/Utils/UnfairQueue/UnfairQueue.swift
ipavlidakis: As we discussed this is because both PRs need the file and i wanted to have on GH and succeed on tests. Once one of them get merged the file should disappear from the other PR

testableapple (approved):

PR #434, opened 2024-06-13
[Enhancement]Participant auto-leave policy implementation

ipavlidakis:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/871

### 🎯 Goal

Ensure that during a ringing-flow call (incoming or outgoing) when the current user become the only one in the call, they will be automatically kicked out.

### 🛠 Implementation

As the coordinator seems to require some time to be fully running after a connection has been recovered from an expired token, rather than relying on the participant_left event from the coordinator, we apply a similar logic when the participants update.

### 🎨 Showcase

|  Before  |  After  |
| -------- | ------- |
|  https://youtu.be/G3Dq48T0ZBk | https://youtu.be/8-bXIMQEdzE |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### ✅ Added
+- A new `ParticipantAutoLeavePolicy` that allows you to set when a user should automatically leave a call. [#434](https://github.com/GetStream/stream-video-swift/pull/434)
+
 ### 🔄 Changed
 
 # [1.0.7](https://github.com/GetStream/stream-video-swift/releases/tag/1.0.7)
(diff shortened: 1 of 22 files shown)


testableapple:
We've QAed this feature with Ilias, looks great ✅
The only tiny issue left is that atm the Picture-In-Picture is stuck when a participant leaves the call ⚠️

martinmitrevski (approved):
looks great 👍  Before merging, we should mention it in the docs and in the changelog.
Also, curious why we need the preconcurrency import for XCTest?

ipavlidakis:
> looks great 👍 Before merging, we should mention it in the docs and in the changelog. Also, curious why we need the preconcurrency import for XCTest?

Adding the docs as we speak. The preconcurrency is due to the usage of `await fulfillment(timeout: defaultTimeout)` as this method expects a Void -> Void closure input and closures aren't sendable and the compiler keeps showing warnings.

PR #447, opened 2024-06-25
[Fix]Flakey tests on LLC and SwiftUI

ipavlidakis:
### 🎯 Goal

Fix flakiness on LLC and SwiftUI tests.

#### LLC
The main reasons for our failing tests seems to be:
1. Updating cancellables storage from multiple threads
2. Using assert methods that are simply waiting for a specified amount of time (i didn't eliminate all of them as there are many, i just fixed the ones failing very often)
3. RTCVideoTracks need to be set to `enabled = false` before deallocation

#### SwiftUI
The main reasons for our failing tests seems to be:
1. Our AudioSession wrapper needs to disable the RTCAudioSession on deinit
2. PeerConnectionFactory needs to cleanUpSSL on deinit (thanks WebRTC for the [docs](https://github.com/ahoyconference/webrtc-ios/blob/master/WebRTC.framework/Headers/RTCSSLAdapter.h). I also did some cleanup for the factory as it seems it had many unnecessary things on it (e.g being an actor when it doesn't maintain any local state, using static methods)
3. Lazily create AudioSession objects
4. Small rewrites on PiPAdapter
5. Make sure that RTCVideoTracks created from Tests are being disabled on tearDown

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideo/Utils/AudioRecorder/StreamCallAudioRecorder.swift
+++ b/Sources/StreamVideo/Utils/AudioRecorder/StreamCallAudioRecorder.swift
@@ -21,7 +21,7 @@ open class StreamCallAudioRecorder: @unchecked Sendable {
     let audioSession: AudioSessionProtocol
 
     /// A private task responsible for setting up the recorder in the background.
-    private var setUpTask: Task<Void, Never>?
+    private var setUpTask: Task<Void, Error>?
 
     private var hasActiveCallCancellable: AnyCancellable?
 
--- a/StreamVideoTests/StreamVideo_Tests.swift
+++ b/StreamVideoTests/StreamVideo_Tests.swift
@@ -143,10 +143,6 @@ final class StreamVideo_Tests: StreamVideoTestCase {
             streamVideo.state.activeCall == nil
                 && streamVideo.state.ringingCall == nil
         }
-
-        // Then
-        XCTAssert(streamVideo.state.ringingCall == nil)
-        XCTAssert(streamVideo.state.activeCall == nil)
     }
     
     func test_streamVideo_incomingCallAccept() async throws {
--- a/StreamVideoTests/WebRTC/WebRTCClient_Tests.swift
+++ b/StreamVideoTests/WebRTC/WebRTCClient_Tests.swift
@@ -40,15 +40,28 @@ final class WebRTCClient_Tests: StreamVideoTestCase {
         participant.name = "Test"
         return participant
     }()
-    
+
     private lazy var participantJoined: Stream_Video_Sfu_Event_ParticipantJoined = {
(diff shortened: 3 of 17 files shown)


StreamVideoTests/StreamVideo_Tests.swift
```diff
@@ -143,10 +143,6 @@ final class StreamVideo_Tests: StreamVideoTestCase {
             streamVideo.state.activeCall == nil
                 && streamVideo.state.ringingCall == nil
         }
-
-        // Then
-        XCTAssert(streamVideo.state.ringingCall == nil)
```
martinmitrevski, line 148: Why is this removed?

Sources/StreamVideo/Utils/AudioRecorder/StreamCallAudioRecorder.swift
```diff
@@ -102,11 +105,30 @@ open class StreamCallAudioRecorder: @unchecked Sendable {
     open func startRecording(ignoreActiveCall: Bool = false) async {
         do {
             let audioRecorder = try await setUpAudioCaptureIfRequired()
-            guard hasActiveCall || ignoreActiveCall, !isRecording else {
-                log
-                    .info(
-                        "🎙️Attempted to start recording but failed. hasActiveCall:\(hasActiveCall) isRecording:\(isRecording)"
-                    )
+            let startRecordingRequest = StartRecordingRequest(
```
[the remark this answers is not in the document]
ipavlidakis: That's right, I moved it here from a cherry pick. Even though it doesn't add change much i'm going to remove it to reduce this PR's scope. Good catch

StreamVideoTests/WebRTC/WebRTCClient_Tests.swift
```diff
@@ -750,10 +784,10 @@ final class WebRTCClient_Tests: StreamVideoTestCase {
         ownCapabilities: [OwnCapability] = [],
         httpClient: HTTPClient? = nil
     ) -> WebRTCClient {
-        let time = VirtualTime()
-        VirtualTimeTimer.time = time
+//        let time = VirtualTime()
```
martinmitrevski, line 787: can we remove the commented out code?

StreamVideoTests/StreamVideo_Tests.swift
ipavlidakis: I have removed both lines below
```swift
XCTAssert(streamVideo.state.ringingCall == nil)
XCTAssert(streamVideo.state.activeCall == nil)
```

because the `fulfilment` implementation above, it performs an XCTAssertTrue for the block passed to it which in our case is
```swift
streamVideo.state.activeCall == nil && streamVideo.state.ringingCall == nil
``` and thus verifies the same thing as the remove XCTAsserts.

martinmitrevski (approved):

PR #496, opened 2024-08-27
[Enhancement]Handle participantsCount event

ipavlidakis:
### 🔗 Issue Links

Resolves https://stream-io.atlassian.net/jira/software/c/projects/PBE/boards/151?selectedIssue=PBE-5756

### 🎯 Goal

Provide a consistent way for participants (regular and anonymous) count observation (before/after joining a call).

### 🛠 Implementation

- Listen and react to the new SFU `call.session_participant_count_updated` event.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideo/CallState.swift
+++ b/Sources/StreamVideo/CallState.swift
@@ -117,6 +117,7 @@ public class CallState: ObservableObject {
     }
 
     @Published public internal(set) var reconnectionStatus = ReconnectionStatus.connected
+    @Published public internal(set) var anonymousParticipantCount: UInt32 = 0
     @Published public internal(set) var participantCount: UInt32 = 0
     @Published public internal(set) var isInitialized: Bool = false
     @Published public internal(set) var callSettings = CallSettings()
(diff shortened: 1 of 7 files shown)


Sources/StreamVideo/CallState.swift
```diff
@@ -117,6 +117,7 @@ public class CallState: ObservableObject {
     }
 
     @Published public internal(set) var reconnectionStatus = ReconnectionStatus.connected
+    @Published public internal(set) var anonymousParticipantCount: UInt32 = 0
```
martinmitrevski, line 120: We should also populate this from `Stream_Video_Sfu_Event_HealthCheckResponse`

Sources/StreamVideo/CallState.swift
```diff
@@ -235,8 +236,12 @@ public class CallState: ObservableObject {
             break
         case .typeCallRtmpBroadcastFailedEvent:
             break
-        case .typeCallSessionParticipantCountsUpdatedEvent:
-            break
+        case let .typeCallSessionParticipantCountsUpdatedEvent(event):
+            participantCount = event.participantsCountByRole
```
martinmitrevski, line 240: can we check if we're not in the call here?

martinmitrevski (approved):

Sources/StreamVideo/CallState.swift
ipavlidakis: I moved the logic to the CallController so we have access on the joined state.

martinmitrevski (approved):

PR #509, opened 2024-09-04
[Fix]Failing tests

ipavlidakis:
### 🎯 Goal

This revision tries to bring most of Tests back online in order to provide status visibility on the reconnection-v2 branch.

### DIFF
--- a/StreamVideoTests/WebRTC/SFU/SFUMiddleware_Tests.swift
+++ b/StreamVideoTests/WebRTC/SFU/SFUMiddleware_Tests.swift
@@ -8,89 +8,89 @@ import XCTest
 
 final class SFUMiddleware_Tests: XCTestCase, @unchecked Sendable {
 
-    private var mockSFUService: MockSignalServer!
-    private var mockSFUAdapter: SFUAdapter!
-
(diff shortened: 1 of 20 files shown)


StreamVideoTests/WebRTC/SFU/SFUMiddleware_Tests.swift
```diff
@@ -8,89 +8,89 @@ import XCTest
 
 final class SFUMiddleware_Tests: XCTestCase, @unchecked Sendable {
 
-    private var mockSFUService: MockSignalServer!
```
martinmitrevski, line 11: shouldn't we just delete this file?

martinmitrevski (approved):

StreamVideoTests/WebRTC/SFU/SFUMiddleware_Tests.swift
ipavlidakis: We will end up removing it, but before doing i would like to write the tests for the new SFUEventAdapter and ensure that all the cases in this file, will be covered in the new tests.

PR #519, opened 2024-09-11
Reconnection v2/fix reconnection when connections lost

ipavlidakis:
### 🔗 Issue Links

_Provide all JIRA tickets and/or GitHub issues related to this PR, if applicable._

### 🎯 Goal

_Describe why we are making this change._

### 📝 Summary

### 🛠 Implementation

### 🎨 Showcase

|  Before  |  After  |
| -------- | ------- |
|  `img`   |  `img`  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### 🎁 Meme


Sources/StreamVideo/Utils/DisposableBag/DisposableBag.swift
```diff
@@ -7,8 +7,9 @@ import Foundation
 
 public final class DisposableBag: @unchecked Sendable {
 
-    private actor Storage {
+    private final class Storage {
```
martinmitrevski, line 10: Curious why we are ditching the actor?

martinmitrevski (approved):

Sources/StreamVideo/Utils/DisposableBag/DisposableBag.swift
ipavlidakis: In the case of the disposableBag it's asynchronicity causes issues. The failing scenario is:
- We are in the JoinedStage and the ws connection drops
- we are transition to the disconnectedStage but before we do we call `removeAll` from the JoinedStage, on the disposableBag
- DisposableBag rather than cancel the encapsulated cancellables immediately it instead spins up a task (in order to interact with the actor)
- In the meanwhile the cancellables in the bag haven't been cancelled (and they are freely running) while the JoinedStage is out of memory and the flow has transitioned further (e.g to Disconnected and then to FastReconnecting because the NWMonitorPath hasn't reported yet that the connection is down)
- While we are in the FastReconnecting stage, the NWMonitorPath sends an update that the connection is down
- The non-cancelled observer from the JoinedStage receives it and triggers a transition to the DisconnectedStage effectively cancelling the fastReconnecting

The last step is partially correct but rather than having an out of memory stage reporting I would like to have the update happening on the active stage (the fast reconnect will fail because the connection is down and thus we will end up again in the disconnected stage).

PR #531, opened 2024-09-19
Add more tests for ICEAdapter

ipavlidakis:
### ☑️ Contributor Checklist

### DIFF
--- a/StreamVideoTests/Mock/MockRTCPeerConnection.swift
+++ b/StreamVideoTests/Mock/MockRTCPeerConnection.swift
@@ -71,7 +71,11 @@ final class MockRTCPeerConnection: StreamRTCPeerConnectionProtocol, Mockable {
 
     // MARK: - Implementation
 
-    var remoteDescription: RTCSessionDescription?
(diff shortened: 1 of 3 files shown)


StreamVideoTests/Mock/MockRTCPeerConnection.swift
```diff
@@ -71,7 +71,11 @@ final class MockRTCPeerConnection: StreamRTCPeerConnectionProtocol, Mockable {
 
     // MARK: - Implementation
 
-    var remoteDescription: RTCSessionDescription?
+    var remoteDescription: RTCSessionDescription? {
+        get { self[dynamicMember: \.remoteDescription] }
```
martinmitrevski, line 75: why do we need this?

ipavlidakis: Just to be aligned with the Mockable protocol. It's not necessary but it can be problematic if we don't do it like this as stubbing the property via Mockable won't have an effect on the actual property.

martinmitrevski (approved):

PR #583, opened 2024-10-22
[Feature]VideoCodec preference

ipavlidakis:
### 🔗 Issue Links

Resolves https://stream-io.atlassian.net/browse/PBE-6265

### 🎯 Goal

Provide a way for integrators, to select their preferred streaming codec.

### 🧪 Manual Testing Notes

- Select the preferred codec from the login screen > debug menu
- Join the call
- Ensure the selected codec is visible in the stats screen next to the Published Resolution stat

### ☑️ Contributor Checklist


DemoApp/Sources/Views/CallView/CallingView/SimpleCallingView.swift
```diff
@@ -161,6 +168,13 @@ struct SimpleCallingView: View {
                 }
         }
     }
+
+    private func setPreferredVideoCodec(for callId: String) async {
```
martinmitrevski, line 172: nit: maybe we can have an extension method defined on the view, since I see this used in a couple of places.

DemoApp/Sources/Views/Login/DebugMenu.swift
```diff
@@ -97,6 +97,10 @@ struct DebugMenu: View {
         didSet { AppEnvironment.autoLeavePolicy = autoLeavePolicy }
     }
 
+    @State private var preferredVideoCodec: AppEnvironment.PreferredVideoCodec = AppEnvironment.preferredVideoCodec {
```
martinmitrevski, line 100: nit: we can remove the type declaration here

Sources/StreamVideo/Call.swift
```diff
@@ -1176,6 +1176,22 @@ public class Call: @unchecked Sendable, WSEventsSubscriber {
         callController.setDisconnectionTimeout(timeout)
     }
 
+    /// Updates video publishing options with the preferred video codec and max bitrate for video
+    /// streaming by.
```
martinmitrevski, line 1180: This feels a bit incomplete, I don't think we need the "by".

Sources/StreamVideo/Controllers/CallController.swift
```diff
@@ -438,10 +438,36 @@ class CallController: @unchecked Sendable {
...
+    ///
+    /// - Parameter timeout: The time interval to set as the timeout.
     func setDisconnectionTimeout(_ timeout: TimeInterval) {
         webRTCCoordinator.setDisconnectionTimeout(timeout)
     }
 
+    // MARK: - Codec preference
+
+    /// Updates video publishing options with the preferred video codec and max bitrate for video
+    /// streaming by.
```
martinmitrevski, line 456: same comment about "by".

Sources/StreamVideo/WebRTC/VideoCapturing/VideoCapturer.swift
```diff
@@ -114,7 +114,7 @@ class VideoCapturer: CameraVideoCapturing {
     }
 
     func updateCaptureQuality(
-        _ codecs: [VideoCodec],
+        _ codecs: [VideoLayer],
```
martinmitrevski, line 117: maybe we can also update the param to layers? (seen it in few places)

Sources/StreamVideo/WebRTC/v2/PeerConnection/Extensions/RTCRtpEncodingParameters+Convenience.swift
```diff
@@ -17,12 +17,19 @@ extension RTCRtpEncodingParameters {
...
+    ) {
         self.init()
-        rid = codec.quality
-        maxBitrateBps = (codec.maxBitrate) as NSNumber
-        if let scaleDownFactor = codec.scaleDownFactor {
-            scaleResolutionDownBy = (scaleDownFactor) as NSNumber
+        rid = layer.quality.rawValue
+        maxBitrateBps = (layer.maxBitrate) as NSNumber
+        if preferredVideoCodec?.isSVC == true {
+            scalabilityMode = "L3T2_KEY"
```
martinmitrevski, line 28: what's this? Some value we have to send for the scalability mode?

Sources/StreamVideo/WebRTC/v2/PeerConnection/MediaAdapters/LocalMediaAdapters/LocalVideoMediaAdapter.swift
```diff
@@ -285,50 +263,123 @@ final class LocalVideoMediaAdapter: LocalMediaAdapting, @unchecked Sendable {
...
+            } else {
+                return false
+            }
+        }()
+        var updatedEncodings = [RTCRtpEncodingParameters]()
+
+        for encoding in params.encodings {
+            let layerSettings = isUsingSVCCodec
+                // for SVC, we only have one layer (q) and often rid is omitted
+                ? layerSettings[0]
```
martinmitrevski, line 301: this is a bit unsafe - can we use `first` instead?

martinmitrevski (commented):
looks good to me! Left couple of comments, but mostly small things.

DemoApp/Sources/Views/CallView/CallingView/SimpleCallingView.swift
ipavlidakis: I'm not against it but my argument is that this will "pollute" the View's API for something that is being used only in 2 places in the Demo App.

Sources/StreamVideo/WebRTC/v2/PeerConnection/Extensions/RTCRtpEncodingParameters+Convenience.swift
ipavlidakis: Yeah, this is straight from the [Spec](https://www.notion.so/stream-wiki/SVC-Codecs-VP9-AV1-and-Dynamic-Simulcast-layer-toggling-1226a5d7f9f680168677c54e1d547600?pvs=4#1226a5d7f9f680acb353f36dfd0ba951). This is starting scalabilityMode we set (3 layers 2 Tracks with Keyframe). However, later on the SFU can send any value from [here](https://www.w3.org/TR/webrtc-svc/#scalabilitymodes*). I thought to make an enum for type-safety but it seems unneccessary. WDYT?

Sources/StreamVideo/WebRTC/v2/PeerConnection/MediaAdapters/LocalMediaAdapters/LocalVideoMediaAdapter.swift
ipavlidakis: Yeap no issues on that. Worth mentioning that in the **current** code [0] is safe, as just above we have checked and ensured that the array isn't empty. As this is only a safety for the current code it makes absolute sense to be a little bit more abstract (and allow optionality) rather than risk having that being a bug on a future refactor.

DemoApp/Sources/Views/CallView/CallingView/SimpleCallingView.swift
martinmitrevski: ok, fair enough

Sources/StreamVideo/WebRTC/v2/PeerConnection/Extensions/RTCRtpEncodingParameters+Convenience.swift
martinmitrevski: no, that's fine, it was just out of curiosity.

martinmitrevski (approved):

PR #588, opened 2024-11-01
[Improvement]Locking mechanisms and Atomic property wrapper

ipavlidakis:
### 🎯 Goal

Improve our locking mechanisms (used to ensure thread-safe access) and add tests to verify stability.

### 📝 Summary

This revision is attempting to align our locking mechanisms (under a single interface) while extending the Atomic property wrapper to be configured with the desired mechanism based on the requirements.

### 🛠 Implementation

- A `LockQueueing` protocol that will be used as the identifying interface between our locking mechanisms.
- The existing `UnfairQueue` conforms to the new `LockQueueing` protocol.
- A new `RecursiveQueue` that uses a `NSRecursiveLock` underneath and conforms to `LockQueueing`.
- An update on the `Atomic` property wrapper that can now be configured with the desired mechanism.
- Added tests that ensure concurrentAccess and deadlock impossibility.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideo/Utils/Queues/LockQueuing.swift
+++ b/Sources/StreamVideo/Utils/Queues/LockQueuing.swift
@@ -6,9 +6,9 @@ import Foundation
 
 /// A protocol defining a lock-based synchronization interface.
 ///
-/// Types conforming to `LockQueing` provide thread-safe access to resources
+/// Types conforming to `LockQueuing` provide thread-safe access to resources
 /// by executing blocks of code within a lock, ensuring mutual exclusion.
-protocol LockQueing {
+protocol LockQueuing {
 
     /// Executes a block within a lock, ensuring exclusive access.
     ///
--- a/Sources/StreamVideo/Utils/Queues/LockQueuing.swift
+++ b/Sources/StreamVideo/Utils/Queues/LockQueuing.swift
@@ -1,0 +1,22 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A protocol defining a lock-based synchronization interface.
+///
+/// Types conforming to `LockQueing` provide thread-safe access to resources
+/// by executing blocks of code within a lock, ensuring mutual exclusion.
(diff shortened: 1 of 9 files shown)


Sources/StreamVideo/Utils/Queues/LockQueuing.swift
```diff
@@ -0,0 +1,22 @@
...
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A protocol defining a lock-based synchronization interface.
+///
+/// Types conforming to `LockQueing` provide thread-safe access to resources
+/// by executing blocks of code within a lock, ensuring mutual exclusion.
+protocol LockQueing {
```
martinmitrevski, line 11: Shouldn't it be LockQueuing?

Sources/StreamVideo/WebSockets/Client/RetryStrategy.swift
```diff
@@ -43,28 +43,25 @@ struct DefaultRetryStrategy: RetryStrategy {
     @Atomic private(set) var consecutiveFailuresCount = 0
     
     mutating func incrementConsecutiveFailures() {
-        _consecutiveFailuresCount.mutate { $0 += 1 }
+        _consecutiveFailuresCount.mutate { $0 + 1 }
     }
     
     mutating func resetConsecutiveFailures() {
-        _consecutiveFailuresCount.mutate { $0 = 0 }
+        consecutiveFailuresCount = 0
```
martinmitrevski, line 50: why do we change this?

martinmitrevski (commented):
Really nice abstraction 👍 We just have a typo with the protocol name, otherwise it's good to merge.

Sources/StreamVideo/WebSockets/Client/RetryStrategy.swift
ipavlidakis: Because the updated property wrapper doesn't require the mutate function(at least directly). Instead, the setter we are calling here will call `mutate` as expected.

PR #589, opened 2024-11-01
[Improvement]SpeakerManager callSettings observation

ipavlidakis:
### 📝 Summary

This revision provides a connecting mechanism that allows a Call's CallSettings observation from SpeakerManager. That will important, in order to allow the SpeakerManager to be updated with speaker/AudioSession updates that haven't been initiated from the SDK's components(e.g. CallKit, Bluetooth headphones connected etc.)

### ☑️ Contributor Checklist


Sources/StreamVideo/Controllers/CallController.swift
```diff
@@ -29,7 +29,7 @@ class CallController: @unchecked Sendable {
         }
     }
 
-    weak var call: Call? {
+    @PublishedWeak var call: Call? {
```
martinmitrevski, line 32: Feels a bit strange to introduce published to a non-observable class. I'm not sure we need the wrapper, let's discuss it a bit more.

ipavlidakis: As we discussed, we are going to remove this one as it doesn't add much value. Thanks

PR #639, opened 2025-01-22
[Enhancement]Improve AudioSession management

ipavlidakis:
### 🔗 Issue Links

Tickets: [[1](https://linear.app/stream/issue/IOS-646/[video][patreon]audio-outputsource-switching-issues), [2](https://linear.app/stream/issue/IOS-522/it-is-not-possible-to-turn-the-sound-off-using-iphone-volume-buttons)]

### 🎯 Goal

Provide a way for users to move audio volume to 0 using the device's physical buttons or the ControlCentre.

### 📝 Summary

We are now driving AudioSession configurations based on CallSettings and OwnCapabilities. However, due to the way that AVAudioSession stores volume profiles per category, we are forced to introduce a way to customise the functionality based on the use-cases. Spefically:
- For calls where all participants are expected to be active it's recommended to do nothing and allow the SDK to use its default behaviour which doesn't allow the user to completely change volume to 0. (this is the expected behaviour iOS has for VoIP calls)
- For livestream or calls where not all participants are expected to be active (some may be just viewers) it's recommended to only change the configuration for every `Viewer`, in your code like below:
```swift
try await call.updateAudioSessionPolicy(OwnCapabilitiesAudioSessionPolicy())
```

### 🛠 Implementation

During callSettings and ownCapabilities changes, the internal AudioSession component will ask the provided `AudioSessionPolicy` to provide the desired AVAudioSession configuration. It's expected that the configuration returned from the `AudioSessionPolicy` should be a valid change. Each policy works differently depending on the input CallSettings and OwnCapabilities:

- `DefaultAudioSessionPolicy`: 
This policy is the one that `Call` objects use by default. The policy will always return the following configuration: category: `.playAndRecord` and categoryOptions:  .allowBluetooth, .allowBluetoothA2DP, .allowAirPlay. While for mode it will toggle between `.videoChat` and `.voiceChat` depending on the `speakerOn` value on CallSettings. **This policy doesn't allow zeroing the volume.**

- `OwnCapabilitiesAudioSessionPolicy`: 
The policy relies on both the provide `CallSettings` and `OwnCapabilities` to produce a valid AudioSessionConfiguration. The rules examined are the following:
- if the user **has no capability to `sendAudio`** it always returns **category: `.playback`, mode: `.default`, categoryOptions: []**, ignoring the `CallSettings`.
- If the user **has capability to `sendAudio`**:
- - if the `CallSettings.audioOn == true` or (`CallSettings.speakerOn == true` and the device is an iPhone): It returns `category:. playAndRecord`. Mode and and CategoryOptions are being calculated similarly to the `DefaultAudioSessionPolicy`.
- In any other case it returns: `category: .playback` similarly on what it would return if the user had no capability to `sendAudio`.

### 🧪 Manual Testing Notes

- You can use the `AudioSession Policy` debug option to try different policies

### ☑️ Contributor Checklist


Sources/StreamVideo/Utils/AudioSession/Policies/DefaultAudioSessionPolicy.swift
```diff
@@ -0,0 +1,31 @@
...
+    ///   - callSettings: The current call settings.
+    ///   - ownCapabilities: The set of the user's own audio capabilities.
+    /// - Returns: The audio session configuration.
+    public func configuration(
+        for callSettings: CallSettings,
+        ownCapabilities: Set<OwnCapability>
+    ) -> AudioSessionConfiguration {
+        .init(
+            category: .playAndRecord,
+            mode: callSettings.speakerOn ? .videoChat : .voiceChat,
```
martinmitrevski, line 26: so if I have video, but no speaker on, what would happen?

ipavlidakis: It will use earpiece. This is because having video doesn't enforce speaker

PR #640, opened 2025-01-23
[Fix]Accepting call on another device will sync state with CallViewModel

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-673/[support]user-accepts-ringing-call-in-one-device-in-app-incoming

### 📝 Summary

When a user accepts a ringing call from one device, on all their other devices correctly dismiss CallKit push notification but they do not update CallViewModel's internal state, resulting in the in-app incoming call to remain on screen.

### 🧪 Manual Testing Notes

- Have a user login on two different devices (and keep the app in the foreground)
- Have another user ring the user from the first step
- Both devices should present CallKit push notification and in-app incoming call
- Accept the call on one device
- The other device, should dismiss CallKit push notification and in-app incoming call screen.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideoSwiftUI/CallViewModel.swift
+++ b/Sources/StreamVideoSwiftUI/CallViewModel.swift
@@ -750,8 +750,18 @@ open class CallViewModel: ObservableObject {
         }
 
         switch callingState {
-        case .incoming where event.user?.id == streamVideo.user.id:
-            setActiveCall(call)
+        case let .incoming(incomingCall)
+            where event.callCid == callCid(from: incomingCall.id, callType: incomingCall.type) && event.user?.id == streamVideo.user
+            .id:
+            /// If the call that was accepted is the incoming call we are presenting, then we reject
+            /// and set the activeCall to the current one in order to reset the callingState to
+            /// inCall or idle.
+            Task {
+                _ = try? await streamVideo
+                    .call(callType: incomingCall.type, callId: incomingCall.id)
+                    .reject()
+                setActiveCall(call)
+            }
         case .outgoing where call?.cId == event.callCid:
             enterCall(
                 call: call,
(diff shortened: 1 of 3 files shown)


Sources/StreamVideoSwiftUI/CallViewModel.swift
```diff
@@ -751,7 +751,7 @@ open class CallViewModel: ObservableObject {
 
         switch callingState {
         case .incoming where event.user?.id == streamVideo.user.id:
-            break
+            rejectCall(callType: event.type, callId: event.callId)
```
martinmitrevski, line 754: Do we really need to explicitly reject the call here? Shouldn't we just update the state to idle? Thinking if this can throw an error, sth like call was already accepted?

ipavlidakis: That's a good point! Updating now

Sources/StreamVideoSwiftUI/CallViewModel.swift
```diff
@@ -750,8 +750,18 @@ open class CallViewModel: ObservableObject {
         }
 
         switch callingState {
-        case .incoming where event.user?.id == streamVideo.user.id:
-            break
+        case let .incoming(incomingCall)
+            where event.callCid == callCid(from: incomingCall.id, callType: incomingCall.type) && event.user?.id == streamVideo.user
```
ipavlidakis, line 754: I simply aligned the implementation with how we are doing it on reject

martinmitrevski (approved):

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: We reverted this change after discussion, as firing `reject` seems to be handle correctly on backend and in that way we align with other SDKs too

PR #641, opened 2025-01-24
[Enhancement]Handle video property on VoIP push notification

ipavlidakis:
### 📝 Summary

By handling the `video` property on the VoIP push notification we can inform CallKit so the call will reported with correct suffix (video/audio)

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideo/Call.swift
+++ b/Sources/StreamVideo/Call.swift
@@ -225,7 +225,7 @@ public class Call: @unchecked Sendable, WSEventsSubscriber {
     ///   - maxDuration: An optional integer representing the maximum duration of the call in seconds.
     ///   - maxParticipants: An optional integer representing the maximum number of participants allowed in the call.
     ///   - backstage: An optional backstage request.
-    ///   - hasVideo: A boolean indicating if the call will be video or only audio. Still requires appropriate
+    ///   - video: A boolean indicating if the call will be video or only audio. Still requires appropriate
     ///   setting of ``CallSettings`.`
     /// - Returns: A `CallResponse` object representing the created call.
     /// - Throws: An error if the call creation fails.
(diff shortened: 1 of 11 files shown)


Sources/StreamVideo/Call.swift
```diff
@@ -238,7 +240,8 @@ public class Call: @unchecked Sendable, WSEventsSubscriber {
         notify: Bool = false,
         maxDuration: Int? = nil,
         maxParticipants: Int? = nil,
-        backstage: BackstageSettingsRequest? = nil
+        backstage: BackstageSettingsRequest? = nil,
+        hasVideo: Bool? = nil
```
martinmitrevski, line 244: I think we should use just `video`, to be inline with the API and the other flags (although I agree `hasVideo` is more descriptive)

Sources/StreamVideoSwiftUI/CallViewModel.swift
```diff
@@ -328,6 +328,8 @@ open class CallViewModel: ObservableObject {
     ///  - maxParticipants: An optional integer representing the maximum number of participants allowed in the call.
     ///  - startsAt: An optional date when the call starts.
     ///  - backstage: An optional request for setting up backstage.
+    ///  - hasVideo: A boolean indicating if the call will be video or only audio. Still requires appropriate
```
martinmitrevski, line 331: This is not enough - we should also put this in the `IncomingCall` model, so customers can provide a different UI based on it. This should be added in `CallEventsHandler.checkForCallEvents`.

ipavlidakis: Good point!

Sources/StreamVideoSwiftUI/Models/IncomingCall.swift
```diff
@@ -17,18 +17,21 @@ public struct IncomingCall: Identifiable, Sendable, Equatable {
...
+    public let video: Bool
 
     public init(
         id: String,
         caller: User,
         type: String,
         members: [Member],
-        timeout: TimeInterval
+        timeout: TimeInterval,
+        video: Bool
```
martinmitrevski, line 28: Should we add a default value (to false) here, since it's a public init?

ipavlidakis: The other values don't have though and it will make the parameter look "less" important. Even if people are using it it will be a very small number of them affected. Up to you.

martinmitrevski (approved):

PR #689, opened 2025-02-25
[Swift6 Migration]Fix Xcode 15 issues

ipavlidakis:
### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/Sources/Components/DemoAppViewFactory.swift
+++ b/DemoApp/Sources/Components/DemoAppViewFactory.swift
@@ -98,7 +98,7 @@ final class DemoAppViewFactory: ViewFactory {
                 onChangeTrackVisibility: onChangeTrackVisibility
             )
             .snapshot(trigger: snapshotTrigger) { [weak viewModel] snapshot in
-                Task { @MainActor in
+                Task { @MainActor [weak viewModel] in
                     viewModel?.sendSnapshot(snapshot)
                 }
             }
--- a/Sources/StreamVideo/Utils/DeviceOrientation/StreamDeviceOrientationAdapter.swift
+++ b/Sources/StreamVideo/Utils/DeviceOrientation/StreamDeviceOrientationAdapter.swift
@@ -120,8 +120,9 @@ open class StreamDeviceOrientationAdapter: ObservableObject, @unchecked Sendable
             // Subscribe to orientation change notifications on UIKit platforms.
             notificationCancellable = notificationCenter
                 .publisher(for: UIDevice.orientationDidChangeNotification)
+                .map { _ in }
                 .receive(on: DispatchQueue.main)
-                .sinkTask { @MainActor [weak self] _ in
+                .sinkTask { @MainActor [weak self] in
                     guard let self = self else { return }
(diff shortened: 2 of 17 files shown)


martinmitrevski (commented):
It's pointing directly to develop

Sources/StreamVideo/Utils/DeviceOrientation/StreamDeviceOrientationAdapter.swift
```diff
@@ -120,8 +120,9 @@ open class StreamDeviceOrientationAdapter: ObservableObject, @unchecked Sendable
             // Subscribe to orientation change notifications on UIKit platforms.
             notificationCancellable = notificationCenter
                 .publisher(for: UIDevice.orientationDidChangeNotification)
+                .map { _ in }
```
martinmitrevski, line 123: why is this needed?

DemoApp/Sources/Components/DemoAppViewFactory.swift
```diff
@@ -98,7 +98,7 @@ final class DemoAppViewFactory: ViewFactory {
                 onChangeTrackVisibility: onChangeTrackVisibility
             )
             .snapshot(trigger: snapshotTrigger) { [weak viewModel] snapshot in
```
martinmitrevski, line 100: do we still need it here?

Sources/StreamVideoSwiftUI/CallingViews/iOS13/BackportStateObject.swift
```diff
@@ -41,10 +41,12 @@ public struct BackportStateObject<ObjectType: ObservableObject & Sendable>: @pre
         self.thunk = thunk
     }
 
-    public mutating func update() {
-        // Not sure what this does but we'll just forward it
-        _state.update()
-        _observedObject.update()
+    nonisolated public func update() {
```
martinmitrevski, line 44: probably good to test this on iOS 13? 😭

DemoApp/Sources/Components/DemoAppViewFactory.swift
ipavlidakis: If we don't add it as weak in the external closure the model will be captured strongly. It's not producing a warning but it's not correct to strongly capture a reference type in a closure.

Sources/StreamVideoSwiftUI/CallingViews/iOS13/BackportStateObject.swift
ipavlidakis: I don't have a way to test this. Xcode doesn't allow me to download any version below iOS 15 and i don't own a device on that version.

ipavlidakis: So, by disabling the @available and forcing the use of the PrejoiningView_iOS13 i'm able to run the app and join a call on both Xcode 15 and 16 when compiling the app for iOS 15.5 (the lowest i can do)

martinmitrevski (approved):

PR #707, opened 2025-03-12
[Improvement]Handle blocked users

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-740/[video]handle-blocked-user-correctly

### 📝 Summary

By default, blocked users are being kicked out from the call (by the backend). If a user somehow manage to find themselves in the call again, we add an additional check that is based on the blockIds from the joinResponse. If the user id is in there, we are making them leave the call.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideo/Controllers/CallController.swift
+++ b/Sources/StreamVideo/Controllers/CallController.swift
@@ -30,7 +30,10 @@ class CallController: @unchecked Sendable {
     }
 
     weak var call: Call? {
-        didSet { subscribeToParticipantsCountUpdatesEvent(call) }
+        didSet {
+            subscribeToParticipantsCountUpdatesEvent(call)
+            subscribeToCurrentUserBlockedState(call)
+        }
     }
 
     private let user: User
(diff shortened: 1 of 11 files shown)


Sources/StreamVideo/Controllers/CallController.swift
```diff
@@ -626,6 +630,39 @@ class CallController: @unchecked Sendable {
         }
     }
 
+    private func subscribeToCurrentUserBlockedState(_ call: Call?) {
+        disposableBag.remove("current-user-blocked")
```
martinmitrevski, line 634: why do we need this?

Sources/StreamVideo/Errors/Errors.swift
```diff
@@ -4,8 +4,6 @@
 
 import Foundation
 
-extension APIError: Error {}
```
martinmitrevski, line 7: We can't remove this unless we change the code generation logic.

ipavlidakis: Very good point! I have forgotten this is generated

Sources/StreamVideo/Controllers/CallController.swift
ipavlidakis: This is to remove the cancellable for the specific observation.

martinmitrevski (approved):

PR #716, opened 2025-03-17
[Fix]InApp screensharing

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-748/[video]inapp-screensharing-issue

### 📝 Summary

It seems that Picture-in-Picture and InApp screenSharing are mutually exclusive. For that reason we fix the InApp screenSharing and provide some logic inside CallViewModel that will avoid breaking the InApp screenSharing flow, in the future.

### 🛠 Implementation

When the user starts an inApp screenShare, we are going to disable Picture-in-Picture and print a warning. 

### 🧪 Manual Testing Notes

- Join a call
- Start inApp screenSharing
- Stop
- Start broadcasting
- Stop
- Start inApp screenSharing
- verify that all tracks work all the time

### ☑️ Contributor Checklist


Sources/StreamVideoSwiftUI/Utils/PictureInPicture/StreamPictureInPictureController.swift
```diff
@@ -20,29 +20,29 @@ final class StreamPictureInPictureController: NSObject, AVPictureInPictureContro
 
     /// The RTCVideoTrack for which the picture-in-picture session is created.
     @MainActor
-    public var track: RTCVideoTrack? {
+    var track: RTCVideoTrack? {
```
martinmitrevski, line 23: the class wasn't public at all, right?

martinmitrevski (approved):
LGTM ✅

Sources/StreamVideoSwiftUI/Utils/PictureInPicture/StreamPictureInPictureController.swift
ipavlidakis: That's right. The class wasn't public but it had properties and methods that were public. Here i'm just removing the unnecessary modifiers

PR #723, opened 2025-03-20
[Fix]Deactivate PiP when the app is in foreground

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-754/pip-should-not-be-visible-when-user-is-in-foreground

### 📝 Summary

In some cases, the app may come to foreground but the PiP will remain active. This revisions fixes this cases.

### 🛠 Implementation

We already have an app state observation in place, in order to disable PiP when the app comes to foreground. As our observer is firing on WillMoveToForeground, we should add a small delay to allow the app to come to foreground and then to stop PiP.

### 🧪 Manual Testing Notes

- Join a call with someone else who shares video
- Move the app to background
- PiP should be active
- In order to come back to the app tap on its icon rather than on PiP
- PiP should get disabled once the app is in the foreground

### ☑️ Contributor Checklist


CHANGELOG.md
```diff
@@ -5,16 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
-- Fix an issue where SetPublisher request was firing when nothing was published [#706](https://github.com/GetStream/stream-video-swift/pull/706)
-- The VideoRendererView for the localParticipant, will flip **only** the front camera feed, not the back one [#708](https://github.com/GetStream/stream-video-swift/pull/708)
-- Attempt to fix a race condition caused in the MicrophoneChecker. [#718](https://github.com/GetStream/stream-video-swift/pull/718)
-
-### ✅ Added
-- Better handling for blocked users [#707](https://github.com/GetStream/stream-video-swift/pull/707)
-
-### ✅ Added
-- Expose publicly the `StreamPictureInPictureAdapter` so it can be used outside of the `CallViewModel` [#711](https://github.com/GetStream/stream-video-swift/pull/711) 
+- Fix an issue that was keeping the Picture-in-Picture active while the app was in the foreground [#723](https://github.com/GetStream/stream-video-swift/pull/723)
```
martinmitrevski, line 8: we're deleting the existing entries here, probably some merge issue?

martinmitrevski (commented):
the solution is good, but we need to fix the changelog before merging

CHANGELOG.md
ipavlidakis: This is because develop hasn't received yet the merged 1.19.0 release. Once we merge main to develop those changes should go away

PR #741, opened 2025-04-02
[Fix]Fast reconnection flow due to ICE misconfiguration

ipavlidakis:
### 📝 Summary

FastReconnection seems to have been broken due to ICE candidates not being applied correctly for the subscriber.

### 🛠 Implementation

We use this revision as an opportunity to align the fast reconnection process to the latest requirements from the SFU. Namely the changes are:
- During fast reconnection, we only restartICE for the publisher connection. That should trigger a negotiation with iceConstraints.
- During fast reconnection, when sending the joinRequest we always use temporary connections to generate the SDP rather than using any existing ones.
- RestartICE on subscriber connection doesn't invoke the `RTCPeerConnection.restartICE` method but instead sends the `iceRestart` RPC call to the SFU. The SFU then will send a new subscriberOffer that will trigger the flow to update the ice state on the subscriber connection.

### 🧪 Manual Testing Notes

- Join a call
- Disable wifi for a second and enable it again
- You should be connected back and all tracks (local and remote) should be active in a reasonable amount of time
- Start screensharing
- Disable wifi for a second and enable it again
- You should be connected back and all tracks (local and remote) should be active in a reasonable amount of time

### ☑️ Contributor Checklist


Sources/StreamVideo/WebRTC/v2/PeerConnection/Adapters/ICEAdapter.swift
```diff
@@ -252,11 +264,13 @@ actor ICEAdapter: @unchecked Sendable {
             .sinkTask(storeIn: disposableBag) { [weak self] in await self?.trickle($0.candidate) }
             .store(in: disposableBag)
 
-        peerConnection
-            .publisher(eventType: StreamRTCPeerConnection.HasRemoteDescription.self)
-            .log(.debug, subsystems: .iceAdapter)
-            .sinkTask(storeIn: disposableBag) { [weak self] _ in await self?.addTrickledICECandidates() }
-            .store(in: disposableBag)
+        if peerType == .publisher {
```
martinmitrevski, line 267: why do we add this check?

Sources/StreamVideo/WebRTC/v2/PeerConnection/MediaAdapters/LocalMediaAdapters/LocalScreenShareMediaAdapter.swift
```diff
@@ -517,4 +517,20 @@ final class LocalScreenShareMediaAdapter: LocalMediaAdapting, @unchecked Sendabl
...
         )
     }
+
+    /// We only want to register the primaryTrack only when we are about to publish for the first time.
+    private func registerPrimaryTrackIfPossible() {
+        guard !hasRegisteredPrimaryTrack else {
+            return
+        }
+
+        subject.send(
```
martinmitrevski, line 527: can this process fail and we end up not registering the primary track?

Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Joining.swift
```diff
@@ -128,7 +128,6 @@ extension WebRTCCoordinator.StateMachine.Stage {
 
                     if isFastReconnecting {
                         await coordinator.stateAdapter.publisher?.restartICE()
-                        await coordinator.stateAdapter.subscriber?.restartICE()
```
martinmitrevski, line 131: why don't we need to restart ICE on the subscriber?

martinmitrevski (commented):
Looks good 👍  Let's do some testing today and tomorrow before merging it.

Sources/StreamVideo/Utils/SerialActor/SerialActorQueue.swift
```diff
@@ -35,18 +49,22 @@ public final class SerialActorQueue: Sendable {
...
                 try await actor.execute(block)
             } catch {
-                // Log any errors encountered during task execution.
-                log.error(
-                    error,
-                    functionName: functionName,
-                    fileName: file,
-                    lineNumber: line
-                )
+                if error is CancellationError { /* No-op */ }
```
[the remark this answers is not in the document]
martinmitrevski: I meant a debug log, not an error log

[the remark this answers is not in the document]
ipavlidakis: I still don't think it's necessary. This is lowest component for executing the tasks. The only way to get cancelled tasks is to do it explicitely or deallocate the object. In both options the higher level code should log.

Sources/StreamVideo/WebRTC/v2/PeerConnection/Adapters/ICEAdapter.swift
ipavlidakis: From the discussion with Marcelo i understood that the the order of applying the candidates matters (needs to happen before creatingAnswer). However upon checking further that doesn't seem to be the case and it's alright to happen at the same time. So removed the check and manual call and now rely on the automatic appliance

Sources/StreamVideo/WebRTC/v2/PeerConnection/MediaAdapters/LocalMediaAdapters/LocalScreenShareMediaAdapter.swift
ipavlidakis: The only thing it does is to send it to a subject, so no errors here.

Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Joining.swift
ipavlidakis: Because of fast reconnection there are no changes, so t here is no need to do anything. That being said, if the peerConnection internally loses its connectivity (due to reasons) we are observing it and we are making a restartICE then (that operation happens [here](https://github.com/GetStream/stream-video-swift/blob/ab4e6c0771c2b0d6a9471debe26cbaa3a7e30421/Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator%2BJoined.swift#L421))

martinmitrevski (approved):

Sources/StreamVideo/WebRTC/v2/SFU/SFUEventAdapter.swift
```diff
@@ -302,40 +294,6 @@ final class SFUEventAdapter: @unchecked Sendable {
         }
     }
 
-    /// Handles a JoinResponse event.
-    ///
-    /// - Parameter event: The JoinResponse event to handle.
-    private func handleJoinResponse(
```
ipavlidakis, line 308: We remove the handling from the SFUEventAdapter as the processing of the JoinResponse happens in [WebRTCCoordinator+Joining](https://github.com/GetStream/stream-video-swift/blob/0637c1c8aaaf013c60220a951355a648b53fdb94/Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator%2BJoining.swift#L338). Having it here causes duplicated processing of the event.

PR #747, opened 2025-04-07
[Fix]Reorganize Call state-machine and its transitions

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-780/improve-call-state-machine

### 📝 Summary

The call state machine management was messy and errors and retries weren't handled correctly. We improve that by moving the logic into the individual stages and reduce complexity from the Call object.

### 🧪 Manual Testing Notes

Try joining a few calls while testing scenarios like joining, calling someone, accepting a call or rejecting. Starting a livestream and joining one

### ☑️ Contributor Checklist


Sources/StreamVideo/CallStateMachine/Stages/Call+JoiningStage.swift
```diff
@@ -0,0 +1,160 @@
...
+        /// - To: JoinedStage, ErrorStage
+        ///
+        /// - Parameter previousStage: The stage transitioning from
+        /// - Returns: Self if transition is valid, nil otherwise
+        override func transition(
+            from previousStage: Call.StateMachine.Stage
+        ) -> Self? {
+            switch previousStage.id {
+            case .idle, .accepted, .joining:
+                execute()
```
martinmitrevski, line 59: So we start a new task, and we immediately return the stage. Can you explain a bit more why we do it like this?

Sources/StreamVideo/CallStateMachine/Stages/Call+JoiningStage.swift
```diff
@@ -0,0 +1,160 @@
...
+
+                do {
+                    let response = try await executeJoin(call: call, input: input)
+                    transitionOrError(.joined(context, response: response))
+                } catch {
+                    var input = input
+                    input.currentNumberOfRetries += 1
+
+                    if input.currentNumberOfRetries < input.maxNumberOfRetries {
+                        let delay = UInt64(TimeInterval.random(in: input.retryDelayRange) * 1_000_000_000)
```
martinmitrevski, line 93: why don't we use the retry policies we already have?

Sources/StreamVideo/CallStateMachine/Stages/Call+Stage.swift
```diff
@@ -0,0 +1,119 @@
...
+            }
+
+            struct RejectingInput {
+                var reason: String?
+                var deliverySubject: PassthroughSubject<RejectCallResponse, Error>
+            }
+
+            weak var call: Call?
+
+            var joinInput: JoinInput? = nil
```
martinmitrevski, line 37: Unsure a bit about all these optional variables, we now have everything in this context.

Sources/StreamVideo/Controllers/CallController.swift
```diff
@@ -74,7 +73,6 @@ class CallController: @unchecked Sendable {
         self.callType = callType
         self.apiKey = apiKey
         self.videoConfig = videoConfig
-        sfuReconnectionTime = 30
```
martinmitrevski, line 77: this wasn't used anywhere, right?

Sources/StreamVideo/CallStateMachine/Stages/Call+JoiningStage.swift
ipavlidakis: This is how our StateMachine works. We do that so that we asynchronously make any operation we need while we allow the state machine to transition to the new stage. That ensures that all operations to the StateMachine are synchronous and short running.

Sources/StreamVideo/CallStateMachine/Stages/Call+JoiningStage.swift
ipavlidakis: The retry policies that accept a closure and obscure all internal transitions. I wanted to move the workload to the statemachine and stop using closures. So now our logs are pointing directly how the operation goes and it's easier to track and debug. It's worth mentioning that the retry and delay logic we have here is the same as on the `RetryPolicy.fastAndSimple`.

Sources/StreamVideo/CallStateMachine/Stages/Call+JoiningStage.swift
```diff
@@ -0,0 +1,160 @@
...
+                notify: input.notify
+            )
+
+            if let callSettings = input.callSettings {
+                try Task.checkCancellation()
+
+                await call.state.update(callSettings: callSettings)
+            }
+
+            try Task.checkCancellation()
```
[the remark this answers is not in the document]
ipavlidakis: That's required though for Swift Concurrency as by simply calling `Task.cancel` does nothing to executing code.

Sources/StreamVideo/Controllers/CallController.swift
```diff
@@ -127,13 +118,16 @@ class CallController: @unchecked Sendable {
         ring: Bool = false,
         notify: Bool = false
     ) async throws -> JoinCallResponse {
+        joinCallResponseSubject = .init(nil)
```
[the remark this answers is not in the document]
ipavlidakis: The StateMachine transitions asynchronously from stage to stage. As we are interested specififcally to the transition to the `joinedStage` which is only possible if there is a joinCallResponse, we create a channel of communication which only delivers this response when available. In that way, we don't have to do filtering on the StateMachine transitions (which removes a lot of the complexity we had before). Finally it allows us to do proper error and timeout handling.

Sources/StreamVideo/Utils/SerialActor/SerialActor.swift
```diff
@@ -7,15 +7,15 @@ import Foundation
...
+    private let disposableBag = DisposableBag()
 
     deinit {
+        disposableBag.removeAll()
         previousTask = nil
     }
 
     nonisolated func cancel() {
-        Task {
-            await previousTask?.cancel()
```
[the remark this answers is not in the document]
ipavlidakis: The task is being converted to a cancellable (which is easier to store as it's already type-erased). All tasks are being stored in the disposableBag that once you call removeAll it cancels all of them.

Sources/StreamVideo/Call.swift
```diff
@@ -130,50 +133,36 @@ public class Call: @unchecked Sendable, WSEventsSubscriber {
...
-                            updateCallSettingsManagers(with: updated)
-                            Task { @MainActor [weak self] in
-                                self?.streamVideo.state.activeCall = self
-                            }
-                            return response
-                        })
-                    }
+        try await callOperationSerialQueue.sync { [weak self] in
+            guard let self else {
+                throw ClientError()
```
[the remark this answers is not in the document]
ipavlidakis: The file,function and line information should be enough for us to debug. For integrators i think this information is irrelevant. Happy to add it though if you want.

Sources/StreamVideo/CallStateMachine/Stages/Call+Stage.swift
ipavlidakis: Yes, that was intended. By having things in the context it's easier to follow and allows stages to be isolated.

Sources/StreamVideo/Controllers/CallController.swift
ipavlidakis: That's right. That variable has been unused since we move to the new WebRTC stack where reconnection time is handled [here](https://github.com/GetStream/stream-video-swift/blob/95988c30a1a088679d1fc28972259b0657f2c833/Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator%2BJoining.swift#L377).

Sources/StreamVideo/CallStateMachine/Stages/Call+JoiningStage.swift
martinmitrevski: Yes, but the logic itself is duplicated. The delay logic seems to be a static function in RetryPolicy, maybe we can just use that?

Sources/StreamVideo/CallStateMachine/Stages/Call+Stage.swift
```diff
@@ -0,0 +1,119 @@
...
+                var create: Bool
+                var callSettings: CallSettings?
+                var options: CreateCallOptions?
+                var ring: Bool
+                var notify: Bool
+                var deliverySubject: PassthroughSubject<JoinCallResponse, Error>
+
+                var currentNumberOfRetries = 0
+                var maxNumberOfRetries = 3
+                var retryDelayRange: ClosedRange<TimeInterval> = 0.25...0.5
```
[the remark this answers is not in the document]
martinmitrevski: I meant we already have this in the `RetryPolicy`:
```
static func quickDelay(retries: Int) -> TimeInterval {
    TimeInterval.random(in: 0.25...0.5)
}
```
We need to maintain it in 2 places (and potentially more, if we need delay range somewhere else).

Sources/StreamVideo/CallStateMachine/Stages/Call+Stage.swift
martinmitrevski: Not a huge fan of it, it's unclear whether all values can be `nil` or only one can be non-nil etc. Maybe we should represent this with enums? (Anyway, not a big thing)

Sources/StreamVideo/Controllers/CallController.swift
martinmitrevski: I believe we fixed this already, we should merge develop.

martinmitrevski (commented):
Looks good in general, left few followups. Want to test this branch today, so good if we can merge develop.

ipavlidakis:
> https://github.com/GetStream/stream-video-swift/pull/747#discussion_r2055454318 
Great suggestion, thanks!

> https://github.com/GetStream/stream-video-swift/pull/747#discussion_r2055443660
Updated to use the existing retry policy struct but not the executeTask.

Sources/StreamVideo/CallStateMachine/Stages/Call+Stage.swift
ipavlidakis: Great suggestion! Updated

Sources/StreamVideo/CallStateMachine/Stages/Call+JoiningStage.swift
ipavlidakis: Changed it to reuse the RetryPolicy struct (but not the executeTask that contains)

martinmitrevski (approved):

PR #751, opened 2025-04-09
[Enhancement]Allow resetting local call settings

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-783/[blink]reset-callsetting-on-callviewmodel-after-call-is-being-left

### 📝 Summary

This revision removes the `CallSettings` propagation, between calls, when using the `CallViewModel`.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideoSwiftUI/CallViewModel.swift
+++ b/Sources/StreamVideoSwiftUI/CallViewModel.swift
@@ -74,6 +74,9 @@ open class CallViewModel: ObservableObject {
     @Published public var callingState: CallingState = .idle {
         didSet {
             handleRingingEvents()
+            if callingState == .idle {
+                prepareForReuse()
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


Sources/StreamVideoSwiftUI/CallViewModel.swift
```diff
@@ -175,7 +173,10 @@ open class CallViewModel: ObservableObject {
     private var callEventsSubscriptionTask: Task<Void, Never>?
     private var participantsSortComparators = defaultSortPreset
     private let callEventsHandler = CallEventsHandler()
-    private var localCallSettingsChange = false
+
+    /// The variable is `true` if CallSettings have been set on the CallViewModel instance (directly or indirectly).
+    /// The variable will be reset to `false` when `leaveCall` will be invoked.
+    private(set) var localCallSettingsChange = false
```
martinmitrevski, line 179: Shouldn't this be configurable though? In meeting apps, I would expect the settings to be remembered between call sessions.

ipavlidakis: So, what i have in my mind is that customers so far are going around this problem by providing their own callSettings whenever they join or create a call. So meeting apps can do that. Basically i'm disabling the propagation as it seems to be unexpected.

martinmitrevski (approved):

PR #771, opened 2025-04-24
[Fix]AudioSession disabling speaker when video is off

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-816/[blink]speaker-turns-off-when-video-is-off

### 📝 Summary

A recent change introduce a regression which is causing the speaker to be disabled when video isn't enabled.

### 🛠 Implementation

The issue is being caused from two parts:
- Our AudioSession policies where wrong (regression introduced in the #762)
- #762 before introducing the regression was trying to fix an issue that was stopping the user from disabling the speaker in CallKit's screen.

In this revision we are trying to solve those issues once an for all. 

By reverting the changes from #762, we get back to the original issue. Strangely the reason why CallKit doesn't allow the turning off of the speaker is because we are using an override.  For some reason CallKit considers that override a hard override and doesn't allow toggling it off.

The solution is to use our ApplicationStateAdapter and use the application state as an input variable when determining the AudioSession configuration. When the app is backgrounded we aren't going to apply any overrides (ports or options) rather than keep whatever the AudioSession.Mode has as default.

### 🧪 Manual Testing Notes

- Join a call
- Ensure that video on/off and speaker on/off combinations work as expected
- Leave the call
- Lock your device
- Receive a ringing call
- Accept
- While in the callKit screen ensure you can turn on/off the speaker

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideo/Utils/AudioSession/Extensions/AVAudioSession.CategoryOptions+Convenience.swift
+++ b/Sources/StreamVideo/Utils/AudioSession/Extensions/AVAudioSession.CategoryOptions+Convenience.swift
@@ -8,13 +8,21 @@ import StreamWebRTC
 extension AVAudioSession.CategoryOptions {
 
     /// Category options for play and record.
-    static func playAndRecord(videoOn: Bool, speakerOn: Bool, appIsInForeground: Bool) -> AVAudioSession.CategoryOptions {
+    static func playAndRecord(
+        videoOn: Bool,
+        speakerOn: Bool,
+        appIsInForeground: Bool
+    ) -> AVAudioSession.CategoryOptions {
         var result: AVAudioSession.CategoryOptions = [
             .allowBluetooth,
             .allowBluetoothA2DP,
             .allowAirPlay
         ]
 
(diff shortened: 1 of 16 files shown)


Sources/StreamVideo/Utils/AudioSession/Extensions/AVAudioSession.CategoryOptions+Convenience.swift
```diff
@@ -8,11 +8,27 @@ import StreamWebRTC
...
+            .allowBluetooth,
+            .allowBluetoothA2DP,
+            .allowAirPlay
+        ]
+
+        /// - Note:We only add the `defaultToSpeaker` if the following are true:
+        /// - It's required (speakerOn = true)
+        /// - The app is foregrounded. The reason is that while in CallKit port overrides are being treated
+        /// as hard overrides and stop CallKit Speaker button from allowing the user to toggle it off.
+        if !videoOn, speakerOn, appIsInForeground {
```
martinmitrevski, line 26: Hmm, I don't really understand this, especially the !videoOn check.

martinmitrevski (approved):

PR #829, opened 2025-06-01
[Performance]UI Improvements

ipavlidakis:
### 🎯 Goal

_Describe why we are making this change._

### 📝 Summary

### 🛠 Implementation

### 🎨 Showcase

|  Before  |  After  |
| -------- | ------- |
|  `img`   |  `img`  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### 🎁 Meme


Sources/StreamVideo/Models/CallParticipant.swift
```diff
@@ -123,8 +123,6 @@ public struct CallParticipant: Identifiable, Sendable, Hashable {
             lhs.sessionId == rhs.sessionId &&
             lhs.connectionQuality == rhs.connectionQuality &&
             lhs.joinedAt == rhs.joinedAt &&
-            lhs.audioLevel == rhs.audioLevel &&
```
martinmitrevski, line 126: This will probably have an impact on the AI integration, since we rely on the audio level there.

Sources/StreamVideo/WebRTC/VideoCapturing/StreamVideoCaptureHandler.swift
```diff
@@ -24,7 +24,7 @@ final class StreamVideoCaptureHandler: NSObject, RTCVideoCapturerDelegate {
 
     init(
         source: RTCVideoCapturerDelegate,
-        handleRotation: Bool = true
+        handleRotation: Bool = false
```
martinmitrevski: why is this changed?

Sources/StreamVideo/WebRTC/VideoFilters/VideoFilters.swift
```diff
@@ -26,7 +26,7 @@ open class VideoFilter: @unchecked Sendable {
     public let name: String
 
     /// Filter closure that takes a CIImage as input and returns a filtered CIImage as output.
-    public var filter: (Input) async -> CIImage
+    public var filter: (Input) -> CIImage
```
martinmitrevski: how big is the impact here? I would prefer to have it async tbh.

Sources/StreamVideoSwiftUI/Utils/PublisherSubscriptionView.swift
```diff
@@ -0,0 +1,90 @@
+//
+// Copyright © 2025 Stream.io Inc. All rights reserved.
+//
+
+import Combine
+import SwiftUI
+
+public struct PublisherSubscriptionView<Value: Equatable, Content: View>: View {
```
martinmitrevski, line 30: Can you explain a bit more - how does this view help with performance?

Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
```diff
@@ -17,7 +17,7 @@ final class VideoRendererPool: @unchecked Sendable {
     ///
     /// - Parameter initialCapacity: The initial capacity of the pool (default is 0).
     @MainActor
-    init(initialCapacity: Int = 0) {
+    init(initialCapacity: Int = 5) {
```
martinmitrevski: why do we change this? Also, we should update the text above if we go with this change.

Sources/StreamVideoSwiftUI/CallView/CallView.swift
```diff
@@ -24,51 +27,114 @@ public struct CallView<Factory: ViewFactory>: View {
     }
 
     public var body: some View {
-        VStack {
-            viewFactory
-                .makeCallTopView(viewModel: viewModel)
-                .presentParticipantEventsNotification(viewModel: viewModel)
+        PublisherSubscriptionView(
```
martinmitrevski, line 30: One thing to consider with custom wrappers is that customers would need to know about it. When they build custom views, they will probably just use observed objects directly.

Sources/StreamVideoSwiftUI/CallView/CallTopView.swift
```diff
@@ -81,16 +101,41 @@ public struct CallTopView: View {
...
 }
 
+public struct CurrentUserScreenSharingIndicatorView: View {
+
+    var viewModel: CallViewModel
+    var call: Call?
+    var sharingPopupDismissed: Binding<Bool>
+
+    public var body: some View {
+        PublisherSubscriptionView(
```
martinmitrevski, line 111: If you need more than one value from the VM, you would need to go back to the standard way, right?

Sources/StreamVideo/Models/CallParticipant.swift
ipavlidakis: We can expose equatable including AudioLevels for AI (closer to the View though and not on the model)

Sources/StreamVideo/WebRTC/VideoCapturing/StreamVideoCaptureHandler.swift
ipavlidakis: Rewriting the frames has an impact but I just changed it back as it doesn't work correctly when rotating the device.

Sources/StreamVideo/WebRTC/VideoFilters/VideoFilters.swift
ipavlidakis: That's something that we need to change to synchronous for sure (we are overly engineering something that it's not in our control). The frame generation pipeline is from WebRTC and the rendering is on AVFoundation. We are just in the middle doing stuff. I'm reverting it for now to reduce the amount of changes but this one has a big asterisk over it.

Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
ipavlidakis: Now that rendering is happening in a more controlled way, this is actually seems to have benefits.

Sources/StreamVideoSwiftUI/CallView/CallView.swift
ipavlidakis: It's ok if customers do it. What's important is that we don't :)

Sources/StreamVideoSwiftUI/CallView/CallTopView.swift
ipavlidakis: You can combine publishers

Sources/StreamVideoSwiftUI/Utils/PublisherSubscriptionView.swift
ipavlidakis: Let's also discuss it offline but the summary is here:

1. Avoids unnecessary view recomputation
* PublisherSubscriptionView isolates publisher observation in a lightweight ViewModel and drives only the local subtree through its contentProvider.
* Unlike onReceive placed in .body, which retriggers the full body recomputation every time the publisher emits, this keeps recomputation localized.

2. More efficient than @ObservedObject/@StateObject in deep trees
* @ObservedObject or @StateObject triggers an entire View recomputation even if only a small part of the state changes.
* In contrast, PublisherSubscriptionView only updates the view model’s value, and only that part of the tree inside contentProvider is updated.
* This is crucial in deeply nested views, where updating the parent could result in cascading recomputations.

3. Avoids binding propagation overhead
* Using Binding to propagate state deep into the hierarchy introduces multiple @State or @Binding layers.
* This increases cognitive load, tightly couples components, and leads to rebuilding multiple views unnecessarily when state changes.
* PublisherSubscriptionView avoids this by scoping observation to exactly where it’s needed, minimizing rebuild impact.

PR #836, opened 2025-06-06
Fix 232ms UI hang in call duration timer

sampepose:
### 🔗 Issue Links
*N/A - Performance improvement identified through profiling*

### 🎯 Goal
Fix a 232ms UI hang that occurs every second during active calls, caused by expensive Swift runtime protocol conformance checking when accessing `@Published` properties in a timer callback.

### 📝 Summary
- **Eliminated 232ms/second UI hang** by caching `startedAt` value to avoid `@Published` property access
- **Moved duration timer to background queue** using `DispatchSourceTimer` instead of `Foundation.Timer`
- **Removed unnecessary `@objc updateDuration()` method** by calculating duration inline
- **Zero impact on functionality** - duration updates work exactly the same, just more efficiently

### 🛠 Implementation
The issue was discovered through Instruments profiling, which showed `swift_dynamicCast` and protocol conformance checking taking 232ms when the duration timer accessed the `@Published var startedAt` property.

**Root cause:** Every second, the timer would access `startedAt` through the `@Published` property wrapper, triggering expensive Swift runtime checks for protocol conformances.

Screenshot 2025-06-06 at 11 25 33 AM

**Solution:**
1. **Cache the value**: Added `_cachedStartedAt` to store the date without going through `@Published`
2. **Background timer**: Switched from `Foundation.Timer` (main thread) to `DispatchSourceTimer` (background queue)
3. **Minimal main thread work**: Only the final `duration` update happens on main thread

This approach eliminates both the protocol conformance overhead AND keeps the timer work off the main thread.

### 🎨 Showcase
|  Before  |  After  |
| -------- | ------- |
| 232ms hang every second during calls | Smooth 60fps during calls |
| Timer blocks main thread | Timer runs on background queue |
| `@Published` access in hot path | Cached value access |

### 🧪 Manual Testing Notes
1. Start a video call
2. Verify the call duration timer updates every second
3. Monitor performance with Instruments - no more periodic hangs
4. Verify timer stops correctly when call ends
5. Test timer behavior during reconnection scenarios

### ☑️ Contributor Checklist
- [x] I have signed the [[Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform)](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)

### DIFF
--- a/Sources/StreamVideo/CallState.swift
+++ b/Sources/StreamVideo/CallState.swift
@@ -97,6 +97,7 @@ public class CallState: ObservableObject {
     @Published public internal(set) var startsAt: Date?
     @Published public internal(set) var startedAt: Date? {
         didSet {
+            _cachedStartedAt = startedAt
             setupDurationTimer()
         }
     }
@@ -153,7 +154,9 @@ public class CallState: ObservableObject {
     }
     
     private var localCallSettingsUpdate = false
-    private var durationTimer: Foundation.Timer?
+    private var durationTimer: DispatchSourceTimer?
+    /// Cache to avoid @Published access
+    private var _cachedStartedAt: Date?
 
     /// We mark this one as `nonisolated` to allow us to initialise a state instance without isolation.
     /// That's a safe operation because `MainActor` is only required to ensure that all `@Published`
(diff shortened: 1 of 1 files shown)


ipavlidakis:
Hey @sampepose, thank you for this PR. We are currently in the process of resolving many performance issues (you can take see a sneak peek in the open PRs).

We are now in a state where we have resolved all the hungs and for that reason I'm not going to accept this PR.

I'm going to keep open though as I reminder to check again once the changes have been merged.

Thanks,
Ilias

sampepose:
Great, I see this was addressed in #825!

ipavlidakis:
Hey @sampepose,

It has been a long time since this PR was open and we have changed/improving things in the meantime. 

I'll close this one but feel free to check the latest version of the SDK and let us know if you find issues with it.

Best regards,
Ilias

PR #839, opened 2025-06-08
Add video renderer pool prewarming API

sampepose:
### 🎯 Goal

Eliminate UI blocking when video renderers are first created during video calls, improving user experience by preventing 150-200ms hangs.

### 📝 Summary

- Add `VideoRendererConfiguration.prewarm()` public API for creating video renderers proactively
- Implement yielding mechanism to prevent continuous UI thread blocking during prewarming
- Fix critical thread safety race condition in `ReusePool` initialization
- Add internal pool configuration method with proper async/await handling

### 🛠 Implementation

**New Public API:**
```swift
// In AppDelegate or early app lifecycle - simple, never blocks
VideoRendererConfiguration.prewarm(count: 4)
```

**Key Implementation Details:**
- **VideoRendererConfiguration**: New public enum providing `prewarm(count:)` method that can be called from any thread
- **Yielding Strategy**: Creates video renderers one at a time with `await Task.yield()` between each creation to prevent blocking the UI thread for extended periods
- **Thread Safety Fix**: Fixed race condition in `ReusePool.init()` where initial elements were created outside of the thread-safe `queue.sync` block
- **Pool Management**: Added `addToAvailable()` method to `ReusePool` for adding pre-created elements beyond initial capacity

**Technical Approach:**
- Video renderers must be created on the main thread (UIView requirement), but prewarming spreads the work across multiple runloop cycles
- Default pool capacity remains 0 (no behavior change without explicit prewarming)
- Frame size during creation is irrelevant as it gets overwritten when renderers are acquired for actual use

### 🧪 Manual Testing Notes

1. **Without Prewarming**: Create a video call and observe ~150-200ms delay when first video view appears
2. **With Prewarming**: Call `VideoRendererConfiguration.prewarm(count: 2)` early in app lifecycle, then create video call - should see immediate video appearance
3. **UI Responsiveness**: During prewarming, verify UI remains responsive (no continuous blocking)
4. **Thread Safety**: Call prewarming from background threads to ensure no crashes

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideoSwiftUI/Utils/ReusePool.swift
+++ b/Sources/StreamVideoSwiftUI/Utils/ReusePool.swift
@@ -30,10 +30,12 @@ final class ReusePool<Element: AnyObject & Hashable> {
         self.initialCapacity = initialCapacity
         self.factory = factory
 
-        // Initialize the pool with a set number of elements
-        for _ in 0..<initialCapacity {
-            let element = factory()
-            available.append(element)
+        // Initialize the pool with a set number of elements (thread-safe)
+        queue.sync {
+            for _ in 0..<initialCapacity {
+                let element = factory()
+                available.append(element)
+            }
         }
     }
 
@@ -92,6 +94,17 @@ final class ReusePool<Element: AnyObject & Hashable> {
                 log.debug("Will release \(inUse.count) \(String(describing: type(of: Element.self))) instances.")
                 inUse.removeAll()
             }
+        }
+    }
+
+    /// Adds a pre-created element to the available pool.
+    /// Used for prewarming to add elements beyond initial capacity.
+    ///
+    /// - Parameter element: The pre-created element to add to the pool.
+    func addToAvailable(_ element: Element) {
+        queue.sync {
+            available.append(element)
+            log.debug("Added prewarmed \(type(of: element)):\(String(describing: element)) to available pool.")
         }
     }
 }
--- a/Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
+++ b/Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
@@ -47,6 +47,25 @@ final class VideoRendererPool: @unchecked Sendable {
     func releaseRenderer(_ renderer: VideoRenderer) {
         pool.release(renderer)
     }
+
+    /// Configures the global video renderer pool with the specified initial capacity.
+    /// This method is intended for internal use by StreamVideo.
+    ///
+    /// - Parameter initialCapacity: The number of video renderers to pre-create.
+    @MainActor
+    static func configure(initialCapacity: Int) async {
+        // Create pool with 0 initial capacity first
+        currentValue = VideoRendererPool(initialCapacity: 0)
+        
+        // Then add renderers one by one with yielding
+        for _ in 0..<initialCapacity {
+            let renderer = VideoRenderer(frame: CGRect(origin: .zero, size: .zero))
+            currentValue.pool.addToAvailable(renderer)
+            
+            // Yield to allow other main thread work between renderer creations
+            await Task.yield()
+        }
+    }
 }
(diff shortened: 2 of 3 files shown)


Sources/StreamVideoSwiftUI/Utils/ReusePool.swift
```diff
@@ -94,4 +96,15 @@ final class ReusePool<Element: AnyObject & Hashable> {
             }
         }
     }
+
+    /// Adds a pre-created element to the available pool.
+    /// Used for prewarming to add elements beyond initial capacity.
+    ///
+    /// - Parameter element: The pre-created element to add to the pool.
+    func addToAvailable(_ element: Element) {
```
ipavlidakis, line 104: This method defies the Pool's purpose. The pool is responsible for creating its elements on demand.

Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
```diff
@@ -47,6 +47,25 @@ final class VideoRendererPool: @unchecked Sendable {
     func releaseRenderer(_ renderer: VideoRenderer) {
         pool.release(renderer)
     }
+
+    /// Configures the global video renderer pool with the specified initial capacity.
+    /// This method is intended for internal use by StreamVideo.
+    ///
+    /// - Parameter initialCapacity: The number of video renderers to pre-create.
+    @MainActor
+    static func configure(initialCapacity: Int) async {
```
ipavlidakis, line 56: This is what is happening on the reusePool initialisation. I don't see a reason for this method.

Sources/StreamVideoSwiftUI/VideoRendererConfiguration.swift
```diff
@@ -0,0 +1,29 @@
...
+    ///
+    /// ## Usage
+    /// ```swift
+    /// // Early in app lifecycle - simple call, never blocks
+    /// VideoRendererConfiguration.prewarm(count: 5)
+    ///
+    /// // Then later initialize StreamVideo as usual
+    /// let streamVideo = StreamVideo(apiKey: apiKey, user: user, token: token)
+    /// ```
+    public static func prewarm(count: Int = 2) {
```
ipavlidakis, line 24: This method looks unnecessary. Can you elaborate more on why this is needed and why the existing warm up on init isn't sufficient?

Sources/StreamVideoSwiftUI/Utils/ReusePool.swift
```diff
@@ -30,10 +30,12 @@ final class ReusePool<Element: AnyObject & Hashable> {
         self.initialCapacity = initialCapacity
         self.factory = factory
 
-        // Initialize the pool with a set number of elements
-        for _ in 0..<initialCapacity {
-            let element = factory()
-            available.append(element)
+        // Initialize the pool with a set number of elements (thread-safe)
+        queue.sync {
```
ipavlidakis, line 34: Hmm i'm not sure about this one. Even without the sync, there aren't multiple threads accessing the initialiser at the same (if they were they would be creating different instances).

Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift
sampepose: The default value is 0: https://github.com/GetStream/stream-video-swift/blob/develop/Sources/StreamVideoSwiftUI/Utils/VideoRendererPool/VideoRendererPool.swift#L20

This entire PR could be removed if that default value is changed, but I didn't want to change the default behavior for all users unexpectedly. So this PR is just a way to expose that number.

ipavlidakis:
Hey @sampepose,

It has been a long time since this PR was open and we have changed/improving things in the meantime. 

I'll close this one but feel free to check the latest version of the SDK and let us know if you find issues with it.

Best regards,
Ilias

PR #852, opened 2025-06-19
[Fix]improve slow network behaviour

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-947/improve-behavior-in-bad-network-connections

### 📝 Summary

The PR attempts to improve the behaviour in bad-network conditions by:
- Remove unnecessary operations on AudioRecorder. Handling is now moved to the WebRTC layer leaving UI components such as MicrophoneChecker being simple readers of data.
- Increase timeouts for WebRTC operations to allow more time for slow networks to process responses.
- Remove unnecessary update on CallSettings from Participants response. In slow networks that was resulting in the local participant getting muted. The operation itself is unnecessary as the same result is being achieved via Websocket events (e.g. if a remote user mute the local participant a WS event will be delivered).
- We are now configuring the audioSession only just before complete the configuration for the publisher. In that way we ensure that the AudioSession handling is only happening for the active call instance (helpful in cases where multiple callInstances may exist in memory).

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideo/WebRTC/v2/WebRTCStateAdapter.swift
+++ b/Sources/StreamVideo/WebRTC/v2/WebRTCStateAdapter.swift
@@ -514,6 +514,39 @@ actor WebRTCStateAdapter: ObservableObject, StreamAudioSessionAdapterDelegate {
         }
     }
 
+    func updateCallSettings(
+        from event: Stream_Video_Sfu_Event_TrackUnpublished
+    ) {
+        guard
+            event.participant.sessionID == sessionID,
+            event.type == .audio || event.type == .video
+        else {
+            return
+        }
+
+        let participant = event.participant.toCallParticipant()
+
+        let currentCallSettings = self.callSettings
+        let possibleNewCallSettings = {
+            switch event.type {
+            case .audio:
+                return currentCallSettings.withUpdatedAudioState(false)
+            case .video:
+                return currentCallSettings.withUpdatedVideoState(false)
+            default:
+                return currentCallSettings
+            }
+        }()
+
+        guard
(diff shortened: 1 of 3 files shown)


Sources/StreamVideo/WebRTC/v2/WebRTCStateAdapter.swift
```diff
@@ -496,30 +514,6 @@ actor WebRTCStateAdapter: ObservableObject, StreamAudioSessionAdapterDelegate {
         }
     }
 
-    /// Updates the call settings from the participants update.
-    /// - Parameter participants: The participants to update the call settings from.
-    /// - Note: This is used when the localParticipant gets muted remotely by someone else.
-    func updateCallSettingsFromParticipants(_ participants: [CallParticipant]) {
```
martinmitrevski, line 502: Does this mean that remote muting of a user won't work anymore?

Sources/StreamVideoSwiftUI/CallingViews/MicrophoneChecker.swift
```diff
@@ -25,6 +24,11 @@ public final class MicrophoneChecker: ObservableObject {
     ) {
         self.valueLimit = valueLimit
         audioLevels = [Float](repeating: 0.0, count: valueLimit)
+        updateMetersCancellable = audioRecorder
+            .metersPublisher
+            .compactMap { [weak self] in self?.normaliseAndAppend($0) }
```
martinmitrevski, line 29: probably this method needs an update now, since the UI doesn't move smoothly in the lobby

PR #861, opened 2025-06-20
[Enhancement]Improve swift concurrency task lifecycle

ipavlidakis:
### 🎯 Goal

Improve Swift Concurrency Task lifecycle by 1. providing a simple way to manage a Task's lifecycle, 2. avoid keeping Tasks in memory to reduce memory usage.

### ☑️ Contributor Checklist


Sources/StreamVideo/CallSettings/CallSettingsManager.swift
```diff
@@ -37,6 +37,9 @@ extension CallSettingsManager {
 }
 
 actor CallSettingsState {
+    private let executor = DispatchQueueExecutor()
```
martinmitrevski, line 40: do we still need these?

Sources/StreamVideo/Utils/CurrentDevice/CurrentDevice.swift
```diff
@@ -2,6 +2,11 @@
 // Copyright © 2025 Stream.io Inc. All rights reserved.
 //
 
+// swiftlint:disable discourage_task_init
+//
+// Copyright © 2025 Stream.io Inc. All rights reserved.
```
martinmitrevski: the header here is added twice

Sources/StreamVideo/Utils/DeviceOrientation/StreamDeviceOrientationAdapter.swift
```diff
@@ -2,6 +2,11 @@
 // Copyright © 2025 Stream.io Inc. All rights reserved.
 //
 
+// swiftlint:disable discourage_task_init
+//
+// Copyright © 2025 Stream.io Inc. All rights reserved.
```
martinmitrevski: same issue with the header

Sources/StreamVideo/Utils/Logger/Signposting.swift
```diff
@@ -0,0 +1,100 @@
...
+
+public var trace: Signposting {
+    if #available(iOS 15.0, *) {
+        Signposter.shared
+    } else {
+        DummySignposter.shared
+    }
+}
+
+final class DummySignposter: Signposting {
```
martinmitrevski: dummy might be a bit misleading? Maybe just use `NoopSignposter`?

Sources/StreamVideo/Utils/Queues/UnfairQueue.swift
```diff
@@ -38,6 +38,7 @@ public final class UnfairQueue: LockQueuing, @unchecked Sendable {
 
     /// Deinitializes the instance, deallocating the unfair lock.
     deinit {
+        lock.deinitialize(count: 1)
```
martinmitrevski, line 41: why 1?

Sources/StreamVideo/Utils/SerialActorQueue/OperationQueue+TaskOperations.swift
```diff
@@ -0,0 +1,146 @@
...
+                line: line,
+                resultSubject: subject,
+                operation: operation
+            )
+        )
+        return try await subject.nextValue(timeout: timeout)
+    }
+}
+
+private final class TaskOperation<Output: Sendable, Failure: Error>: Operation, @unchecked Sendable {
```
martinmitrevski: This stuff is really awesome! I'm thinking if we should move these at some point to `StreamCore`, and provide building blocks to all products (of course, outside of this PR/line of work). But sometimes we should consider in the next few months.

Sources/StreamVideo/WebRTC/v2/PeerConnection/Adapters/ICEAdapter.swift
```diff
@@ -23,6 +23,9 @@ actor ICEAdapter: @unchecked Sendable {
     private var pendingSFUCandidates: [RTCIceCandidate] = []
     private var pendingLocalCandidates: [RTCIceCandidate] = []
 
+    private let executor = DispatchQueueExecutor()
+//    nonisolated var unownedExecutor: UnownedSerialExecutor { .init(ordinary: executor) }
```
martinmitrevski: I guess we should remove this

Sources/StreamVideo/WebRTC/v2/PeerConnection/RTCTemporaryPeerConnection.swift
```diff
@@ -73,7 +73,9 @@ final class RTCTemporaryPeerConnection {
     /// Cleans up resources when the instance is being deallocated.
     deinit {
         peerConnection.transceivers.forEach { $0.stopInternal() }
+        // swiftlint:disable discourage_task_init
```
martinmitrevski, line 76: maybe we're too strict with this rule?

Sources/StreamVideo/WebSockets/Events/EventNotificationCenter.swift
```diff
@@ -8,7 +8,7 @@ import Foundation
 class EventNotificationCenter: NotificationCenter, @unchecked Sendable {
     private(set) var middlewares: [EventMiddleware] = []
 
-    var eventPostingQueue = DispatchQueue(label: "io.getstream.event-notification-center")
+    var eventPostingQueue = DispatchQueue(label: "io.getstream.event-notification-center", qos: .background)
```
martinmitrevski: not sure if `qos: .background` is suitable here, I would keep the `default` tbh. Can you share your thinking?

Sources/StreamVideoSwiftUI/CallViewModel.swift
```diff
@@ -1030,3 +1054,66 @@ public enum ParticipantsLayout {
     case spotlight
     case fullScreen
 }
+
+final class ParticipantEventResetAdapter: @unchecked Sendable {
```
martinmitrevski: This is also nice, we needed something like this

Sources/StreamVideo/CallSettings/CallSettingsManager.swift
ipavlidakis: The reason for keeping it is that this is adding tasks outside of the Cooperative pool allowing us to run long running tasks without blocking the threads that are being used across the app.

Sources/StreamVideo/Utils/Queues/UnfairQueue.swift
ipavlidakis: Because the allocated capacity was 1, so we are only deinitializing what we allocated intially.

Sources/StreamVideo/Utils/SerialActorQueue/OperationQueue+TaskOperations.swift
ipavlidakis: Definitely!

Sources/StreamVideo/WebRTC/v2/PeerConnection/Adapters/ICEAdapter.swift
ipavlidakis: Actually i'm going to bring it back as i had commented to check something and i forgot to reenable it.

Sources/StreamVideo/WebRTC/v2/PeerConnection/RTCTemporaryPeerConnection.swift
ipavlidakis: That's actually intended. The plan is to avoid having Tasks that there is no way to control their lifecycle (e.g. cancel them). Do you think we should relax it?

Sources/StreamVideo/WebSockets/Events/EventNotificationCenter.swift
ipavlidakis: My thinking is to set low qos to things that are running in the background but that may be too aggressive (in the Tasks implementation that was causing issues). I think you suggestion is right, i'm going to change this to `.default`

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: Just moved it to a seperate file

Sources/StreamVideo/WebRTC/v2/PeerConnection/RTCTemporaryPeerConnection.swift
martinmitrevski: I think so - we should have this only in the core parts. What do you think?

Sources/StreamVideo/WebSockets/Events/EventNotificationCenter.swift
martinmitrevski: yes, we need to see these case by case. The notification center should have higher than bg.

Sources/StreamVideo/Utils/Queues/UnfairQueue.swift
martinmitrevski: Yes, but the problem is what if we change something in the future - we always have to remember to update this as well?

martinmitrevski (approved):

PR #888, opened 2025-07-14
[Enhancement]Incoming video pause

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-994/featureinbound-video-pause

### 🎯 Goal

Provide a way for the SFU to instruct the client which tracks are remotely paused.

### 📝 Summary

Docs preview: https://github.com/GetStream/docs-content/pull/452

### 🧪 Manual Testing Notes

- By default the feature is disabled
- Enter a call with another client
- Activate the Link conditioner to Edge
- Remain in the call a few seconds
- You shouldn't observe any changes during the call. Remote track should be lagging but always visible
- Disable the link conditioner
- Logout and enable the feature from the debug menu (Client Capabilities > Subscriber video pause)
- Enter a call with another client
- Activate the Link conditioner to Edge
- Remote track should get disabled at some point (it's ok if it gets reenabled)

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+### ✅ Added
+- `ClientCapabilities` have been added to support remote subscriber track pause. [#888](https://github.com/GetStream/stream-video-swift/pull/888)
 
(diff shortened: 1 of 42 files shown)


ipavlidakis:
> We should have this on by default, without it, it breaks the experience on low-end networks. We can offer an opt out if needed.

There is opt-out as integrators can simply pass an empty array for the clientCapabilities. Let me align that with React too.

PR #906, opened 2025-07-31
[Enhancement]Improved audio session management

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1046/audio-issues

### 🎯 Goal

Provide a consistent and centralised way for managing AudioSession updates across SDK Components and Calls.

### 🛠 Implementation

This revision introduces `RTCAudioStore` which is meant to be the control centre for AudioSession for the SDK, completely abstracting AVAudioSession and RTCAudioSession. The reason for this abstraction is that AVAudioSessionManagement and CallKit require things to happen in serial order. This order is what `RTCAudioStore` offers. 

The architecture is Redux-based with 2 main Reducers (RTCAudioSessionReducer and CallKitAudioSessionReducer). Each of them is responsible with interacting with one external component (in this case RTCAudioSession and CallKit). 

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+### 🐞 Fixed
+- AudioSession management issues that were causing audio not being recorded during calls. [#906](https://github.com/GetStream/stream-video-swift/pull/906)
 
 # [1.29.1](https://github.com/GetStream/stream-video-swift/releases/tag/1.29.1)
 _July 25, 2025_


Sources/StreamVideo/Utils/AudioSession/StreamAudioSession.swift
```diff
@@ -339,76 +419,97 @@ final class StreamAudioSession: @unchecked Sendable, ObservableObject {
...
             }
 
             do {
                 try await audioSession.setCategory(
-                    configuration.category,
-                    mode: configuration.mode,
-                    with: configuration.options
+                    .playAndRecord,
+                    mode: .voiceChat,
+                    with: .allowBluetooth
```
martinmitrevski: why are these hardcoded now?

ipavlidakis: The StreamAudioSession isn't used any more. I'll remove it.

Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Joined.swift
```diff
@@ -397,11 +398,6 @@ extension WebRTCCoordinator.StateMachine.Stage {
                 .removeDuplicates()
                 .sinkTask(storeIn: disposableBag) { [weak self] callSettings in
                     guard let self else { return }
-
-                    if let statsAdapter = await context.coordinator?.stateAdapter.statsAdapter {
```
martinmitrevski, line 401: why is this removed?

Sources/StreamVideo/WebRTC/v2/WebRTCStateAdapter.swift
```diff
@@ -124,15 +124,7 @@ actor WebRTCStateAdapter: ObservableObject, StreamAudioSessionAdapterDelegate {
         self.rtcPeerConnectionCoordinatorFactory = rtcPeerConnectionCoordinatorFactory
         self.videoCaptureSessionProvider = videoCaptureSessionProvider
         self.screenShareSessionProvider = screenShareSessionProvider
-        self.audioSession = .init(audioDeviceModule: peerConnectionFactory.audioDeviceModule)
```
martinmitrevski, line 127: we should test audio filters and noise cancellation after these changes

Sources/StreamVideo/WebRTC/v2/WebRTCStateAdapter.swift
```diff
@@ -363,6 +363,8 @@ actor WebRTCStateAdapter: ObservableObject, StreamAudioSessionAdapterDelegate {
         set(anonymousCount: 0)
         set(participantPins: [])
         trackStorage.removeAll()
+//        try? await audioSession.deactivate(.internal)
```
martinmitrevski: we should delete this

Sources/StreamVideo/WebRTC/v2/WebRTCStateAdapter.swift
```diff
@@ -634,36 +637,20 @@ actor WebRTCStateAdapter: ObservableObject, StreamAudioSessionAdapterDelegate {
...
-                    log.error(error)
-                }
-            }
-            .store(in: disposableBag)
+    func configureAudioSession(source: JoinSource?) async throws {
+        audioSession.activate(
+            callSettingsPublisher: $callSettings.removeDuplicates().eraseToAnyPublisher(),
+            ownCapabilitiesPublisher: $ownCapabilities.removeDuplicates().eraseToAnyPublisher(),
+            delegate: self,
+            /// If we are joining from CallKit the AudioSession will be activated from it and we
```
martinmitrevski, line 644: this is irrelevant of call settings and whether you are muted initially, right?

Sources/StreamVideo/Call.swift
```diff
@@ -147,6 +147,24 @@ public class Call: @unchecked Sendable, WSEventsSubscriber {
...
+        /// This block checks if the `joinSource` has already been set in the current
+        /// call state. If not, it assigns `.inApp` as the default join source,
+        /// indicating the call was joined from within the app UI. The resolved
+        /// `JoinSource` value is then used to record how the call was joined,
+        /// enabling analytics and behavioral branching based on entry point.
+        let joinSource = await {
+            if let joinSource = await state.joinSource {
+                return joinSource
+            } else {
+                return await Task { @MainActor in
```
martinmitrevski, line 161: why is it so complex? 😅

Sources/StreamVideo/Utils/AudioSession/CallAudioSession.swift
```diff
@@ -0,0 +1,192 @@
...
+                subsystems: .audioSession
+            )
+            return
+        }
+
+        audioStore.dispatch(
+            .audioSession(
+                .setCategory(
+                    .playAndRecord,
+                    mode: .voiceChat,
```
martinmitrevski, line 194: why is it `voiceChat` here?

Sources/StreamVideo/Utils/AudioSession/RTCAudioStore/Effects/RTCAudioStore+InterruptionEffect.swift
```diff
@@ -0,0 +1,96 @@
...
+            _ session: RTCAudioSession,
+            shouldResumeSession: Bool
+        ) {
+            guard let store else {
+                return
+            }
+
+            store.dispatch(.audioSession(.isInterrupted(false)))
+            if shouldResumeSession {
+                store.dispatch(.audioSession(.isActive(false)))
```
martinmitrevski, line 74: as discussed, this is not ideal, since the subsequent actions proceed anyway - dispatch swallows the errors. Good to think about possible edge cases here.

Sources/StreamVideo/Call.swift
ipavlidakis: Thanks to the MainActor limitation on the call state

Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Joined.swift
ipavlidakis: The callSettings and the audioSession were passed to the statsAdapter so that it can report traces whenever there was a change. Now that the CallAudioSession is lighter i moved this functionality over there.

Sources/StreamVideo/WebRTC/v2/WebRTCStateAdapter.swift
ipavlidakis: Audio filters shouldn't be affected as they are running directly in the WebRTC audio device module pipeline (which is the lower lever of the AudioSession we are using). In any case we should test them

Sources/StreamVideo/Utils/AudioSession/CallAudioSession.swift
ipavlidakis: VoiceChat is the middle of the modes (default isn't optimized and video has more optimisations). As the initialConfiguration runs when a call is being created and targets the configuration required for CallKit i have settled for the voiceChat as video isn't possible on CallKit. That being said, when we join the call the fetched callSetting will update this configuration.

Sources/StreamVideo/Utils/AudioSession/RTCAudioStore/Effects/RTCAudioStore+InterruptionEffect.swift
ipavlidakis: The error here wouldn't provide any feedback enough to make an action because an error on setActive doesn't mean that the session isn't active. The only reason setActive may fail is because callKit has the session active which is still valid for the following configurations. 

That being said we should invest into recovery mechanisms for cases where the audioSession has become terribly broken (e.g. rejoins). However i would leave the implementation of those mechanisms for another PR

Sources/StreamVideo/WebRTC/v2/WebRTCStateAdapter.swift
ipavlidakis: You are right. Basically when the user accepts the call through a callkit component (screen or notification) then CallKit activates the audiosession. The difference here is that CallKit activation uses a higher priority configuration which basically makes any call to setActive (while this configuration is active) to fail. The check here is to avoid this error.

martinmitrevski (approved):
Great work! ✅

PR #909, opened 2025-08-11
Fix PiP UI issues on iOS 26

martinmitrevski:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1065/fix-pip-ui-issues-on-ios-26.

### 🎯 Goal

Fix UI issues on PiP on iOS 26.

### 📝 Summary

### 🛠 Implementation

### 🎨 Showcase

|  Before  |  After  |
| -------- | ------- |
|  `img`   |  `img`  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### 🎁 Meme

### DIFF
--- a/Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
+++ b/Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
@@ -413,7 +413,7 @@ public struct ParticipantInfoView: View {
     var maxHeight: CGFloat
     var paddingsConfig: EdgeInsets
 
-    @MainActor public init(
+    public init(
         participant: CallParticipant,
         isPinned: Bool,
         maxHeight: Float = 14,
--- a/Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
+++ b/Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
@@ -411,15 +411,18 @@ public struct ParticipantInfoView: View {
     var participant: CallParticipant
     var isPinned: Bool
     var maxHeight: CGFloat
+    var paddingsConfig: PaddingsConfig
 
     public init(
         participant: CallParticipant,
         isPinned: Bool,
-        maxHeight: Float = 14
+        maxHeight: Float = 14,
+        paddingsConfig: PaddingsConfig = .participantInfoView
     ) {
         self.participant = participant
         self.isPinned = isPinned
         self.maxHeight = CGFloat(maxHeight)
+        self.paddingsConfig = paddingsConfig
     }
     
     public var body: some View {
@@ -452,8 +455,10 @@ public struct ParticipantInfoView: View {
             SoundIndicator(participant: participant)
                 .frame(maxHeight: maxHeight)
         }
-        .padding(.all, 2)
-        .padding(.horizontal, 4)
+        .padding(.leading, paddingsConfig.leading)
+        .padding(.trailing, paddingsConfig.trailing)
+        .padding(.top, paddingsConfig.top)
+        .padding(.bottom, paddingsConfig.bottom)
         .frame(height: 28)
         .cornerRadius(
             8,
--- a/Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
+++ b/Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
@@ -411,13 +411,13 @@ public struct ParticipantInfoView: View {
     var participant: CallParticipant
     var isPinned: Bool
     var maxHeight: CGFloat
-    var paddingsConfig: PaddingsConfig
+    var paddingsConfig: EdgeInsets
 
     @MainActor public init(
         participant: CallParticipant,
         isPinned: Bool,
         maxHeight: Float = 14,
-        paddingsConfig: PaddingsConfig = .participantInfoView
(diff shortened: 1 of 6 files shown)


Sources/StreamVideoSwiftUI/Utils/PaddingsConfig.swift
```diff
@@ -0,0 +1,119 @@
...
+
+extension PaddingsConfig {
+    
+    /// Padding configuration for the participant info view.
+    ///
+    /// - Leading: `6`
+    /// - Trailing: `6`
+    /// - Top: `2`
+    /// - Bottom: `2`
+    @MainActor public static let participantInfoView = PaddingsConfig(
```
ipavlidakis: Why @MainActor? Can we simply make it nonisolated(unsafe)?

Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
```diff
@@ -452,8 +455,10 @@ public struct ParticipantInfoView: View {
             SoundIndicator(participant: participant)
                 .frame(maxHeight: maxHeight)
         }
-        .padding(.all, 2)
-        .padding(.horizontal, 4)
+        .padding(.leading, paddingsConfig.leading)
```
ipavlidakis: Does it make sense to create simple modifier that applies the paddingsConfig directly so we don't have to add 4 lines of padding configuration per view + make sure that always the right value is being applied on the right side?

Sources/StreamVideoSwiftUI/Utils/PaddingsConfig.swift
```diff
@@ -0,0 +1,119 @@
...
+// Copyright © 2025 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A configuration object that defines padding values for different edges.
+///
+/// `PaddingsConfig` allows you to specify custom padding values for leading,
+/// trailing, top, and bottom edges, which can be reused across multiple UI components.
+public struct PaddingsConfig {
```
ipavlidakis: Why not use the EdgeInsets struct which is already defined?

Sources/StreamVideoSwiftUI/Utils/PaddingsConfig.swift
martinmitrevski: all this stuff should be on the main actor

Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
martinmitrevski: that's a good point, will do that

Sources/StreamVideoSwiftUI/Utils/PaddingsConfig.swift
martinmitrevski: it's true that it has the same properties, but it's not really the same purpose. Hmm, not sure, bit on the fence here.

Sources/StreamVideoSwiftUI/Utils/PaddingsConfig.swift
ipavlidakis: Why is that? Those things aren't access concurrently and the @MainActor is currently messing our API

Sources/StreamVideoSwiftUI/Utils/PaddingsConfig.swift
ipavlidakis: What is the difference that gets you on the fence. They look quite similar to me.

ipavlidakis: As a bonus, if you change it to EdgeInsets you don't need to implement the modifier as the existing `.padding` modifier handles `EdgeInsets` already

Sources/StreamVideoSwiftUI/Utils/PaddingsConfig.swift
martinmitrevski: they are only accessed from the main actor in any case

Sources/StreamVideoSwiftUI/Utils/PaddingsConfig.swift
martinmitrevski: it's not really the same as padding: https://developer.apple.com/documentation/swiftui/edgeinsets.
But, the handling of the padding modifier resolves any doubts for me, so I will do the change.

Sources/StreamVideoSwiftUI/Utils/PaddingsConfig.swift
ipavlidakis: The fact they are accessed only from MainThread doesn't require the @MainActor limitation.

martinmitrevski: not needed now that we use edge insets

Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
martinmitrevski: this is also not needed now

Sources/StreamVideoSwiftUI/Utils/PaddingsConfig.swift
martinmitrevski: updated, looks much nicer now

ipavlidakis (approved):

PR #913, opened 2025-08-13
[Fix]CallSettings rapid updates

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-916/videoproximity-doesnt-stop-local-video-and-its-visible-on-remote

### 🎯 Goal

This revisions provides improvements in situations that causing CallSettings to update rapidly. With the updates, we can now ensure serial updates for those rapid changes.

### 🛠 Implementation

`WebRTCStateAdapter` now handles `CallSettings` updates as enqueued operations. In that way every update request will be passed the current CallSettings at the moment and is expected to return the update it needs. In this way we provide step by step updates in CallSettings rather than complete replacements.

### 🧪 Manual Testing Notes

- With proximity policies enabled
- Join call
- Move your hand close to the front camera
- Notice that video and speaker should be disabled
- Move your hand away from the front camera
- Video and Speaker should be reenabled
- Do that multiple times with varying speed
- When you move away from the front camera, video and speaker should always be restored

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Joined.swift
+++ b/Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Joined.swift
@@ -119,10 +119,6 @@ extension WebRTCCoordinator.StateMachine.Stage {
 
                     try Task.checkCancellation()
 
-                    await observeCallSettingsUpdates()
-
-                    try Task.checkCancellation()
-
                     await observePeerConnectionState()
 
                     try Task.checkCancellation()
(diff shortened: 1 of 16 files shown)


Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Joined.swift
```diff
@@ -119,10 +119,6 @@ extension WebRTCCoordinator.StateMachine.Stage {
 
                     try Task.checkCancellation()
 
-                    await observeCallSettingsUpdates()
```
martinmitrevski, line 122: why do we remove this?

Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Joined.swift
```diff
@@ -386,42 +382,6 @@ extension WebRTCCoordinator.StateMachine.Stage {
                 .store(in: disposableBag)
         }
 
-        /// Observes updates to the `callSettings` and ensures that any changes are
-        /// reflected in the publisher. This ensures that updates to audio, video, and
-        /// audio output settings are applied correctly during a WebRTC session.
-        private func observeCallSettingsUpdates() async {
```
martinmitrevski, line 392: I think I haven't seen a replacement for this? How do we listen to updates now?

Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Joined.swift
ipavlidakis: This method was here to forward CallSettings to the `publisher` PeerConnection. However, because it was relying on the Combine chain for updates, it was always a time-fracture behind.

Now, we are removing this and [moving](https://github.com/GetStream/stream-video-swift/blob/876bfe5b60a861f56dc54826a6415926fb9771d1/Sources/StreamVideo/WebRTC/v2/WebRTCStateAdapter.swift#L508) the `publiher` peerconnection update directly on the `WebRTCStateAdapter` next to when it updates its own CallSettings.

PR #914, opened 2025-08-19
[Enhancement]Simplify `Task.timeout` and `Publisher.nextValue`

ipavlidakis:
### 🎯 Goal

Simplify and improve the `Publisher.nextValue` and `Task.timeout` APIs.

### 🛠 Implementation

Reference for `Task.timeout` API changes: https://x.com/twannl/status/1950095104254849246

### 🧪 Manual Testing Notes

- Join multiple calls and ensure joining works as expected.

### ☑️ Contributor Checklist


StreamVideoTests/Mock/MockRTCAudioStore.swift
```diff
@@ -16,6 +16,10 @@ final class MockRTCAudioStore {
         audioStore = RTCAudioStore(session: session)
     }
 
+    deinit {
+        InjectedValues[\.audioStore] = .init()
```
martinmitrevski, line 20: Is this a leftover from the other PR?

martinmitrevski (commented):
This is just a refactoring task, right? I would prefer to review and merge it after the release, since we will put other bigger stuff as well. Wdyt?

StreamVideoTests/Mock/MockRTCAudioStore.swift
ipavlidakis: This one is actually here to clean the InjectedValues context when the store isn't needed any more. Now that i'm thinking out loud though, it will never be triggered as the reference is still in InjectedValues, so deinit will never be triggered. We probably need an explicit method to uninstall/dismantle.

PR #921, opened 2025-08-25
[Enhancement]Local participant videoRender binded on CallSettings

ipavlidakis:
### 📝 Summary

This revision simplifies VideoRenderer configuration by relying on CallSettings rather that customData.

### 🛠 Implementation

Since the beginning of the project we are controlling the VideoRenderer for the LocalParticipant using customData, to decided if we need to show video or not. As the project has evolved, we have a better more consistent way to handle this, CallSettings observation. The CallSettings observation not only aligns the component with the rest of the UI components in the SDK but also fixes the issue where the VideoRenderer for the local participant was "disconnected" from the CallSettings and was solely relying on rerenders to get the correct value to show/hide the video.

### 🧪 Manual Testing Notes

Join a few calls, toggle video on/off ensure that the VideoRenderer for your user is always showing what you are expecting it to.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
+++ b/Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
@@ -2,6 +2,7 @@
 // Copyright © 2025 Stream.io Inc. All rights reserved.
 //
 
+import Combine
 import StreamVideo
 import StreamWebRTC
 import SwiftUI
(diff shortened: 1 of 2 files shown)


Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
```diff
@@ -325,7 +326,10 @@ public struct VideoCallParticipantView<Factory: ViewFactory>: View {
     var customData: [String: RawJSON]
     var call: Call?
 
-    @State private var isUsingFrontCameraForLocalUser: Bool = false
+    private var isLocalParticipant: Bool
+
+    private var callSettingsPublisher: AnyPublisher<CallSettings, Never>?
```
martinmitrevski, line 332: I think we discussed this once - is it safe that this is recreated on every init?

ipavlidakis: I don't see any reason why that should cause issues? Publishers are just reference boxes. They are cheap to create and they only start pushing values when they receive their first subscription. We are safe here.

martinmitrevski (approved):

PR #955, opened 2025-10-01
[CI] Support Xcode 26

testableapple:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1174


StreamVideo.xcodeproj/project.pbxproj
```diff
@@ -10818,6 +10818,7 @@
 				SKIP_INSTALL = YES;
 				SWIFT_EMIT_LOC_STRINGS = YES;
 				SWIFT_OBJC_BRIDGING_HEADER = "";
+				SWIFT_OPTIMIZATION_LEVEL = "-Osize";
```
laevandus: Thinking if we should keep it unchanged since most of the customers use the default setting. This just makes the report nicer.

testableapple: We also propagate this change in Package.swift and Podfiles, so all the customers will use it as well.

ipavlidakis: I don't see why we make this change?

testableapple: To reduce the SDK size

testableapple: I got rid of the flag 👍

laevandus (approved):
✅

PR #967, opened 2025-10-09
[CI] Implement detailed size metrics

testableapple:
Resolve https://linear.app/stream/issue/IOS-1195

Screenshot 2025-10-10 at 4 17 14 PM

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -786,6 +786,7 @@ private_lane :frameworks_sizes do
 
   gym(
     scheme: 'DemoAppUIKit',
+    configuration: 'Release',
     archive_path: archive_dir,
     export_method: 'ad-hoc',
     export_options: 'fastlane/sdk_size_export_options.plist',


StreamVideo.xcodeproj/project.pbxproj
```diff
@@ -9882,6 +9882,8 @@
 				INFOPLIST_KEY_NSHumanReadableCopyright = "";
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
 				IPHONEOS_DEPLOYMENT_TARGET = 13.0;
+				LD_GENERATE_MAP_FILE = YES;
+				LD_MAP_FILE_PATH = "linkmaps/$(PRODUCT_NAME)-$(CURRENT_ARCH)-LinkMap.txt";
```
martinmitrevski, line 9886: is this also needed for the release config?

testableapple: I think in release config we get the most correct numbers, so the calculation is done on release config. However, it does not affect the release sdk at all, it just generates the additional artifact outside of the framework.

PR #988, opened 2025-10-31
[Fix]Mistakenly prompting for permissions

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1209/buglivestream-requesting-permission-to-speak

### 🛠 Implementation

This revision fixes an issue that was causing permission prompts to appear when they weren't needed.

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Fix permission prompts appearing when not needed. [#988](https://github.com/GetStream/stream-video-swift/pull/988)
 
+### ✅ Added
+- SwiftUI modifiers that surface moderation blur and warning events in `CallView`. [#987](https://github.com/GetStream/stream-video-swift/pull/987)
+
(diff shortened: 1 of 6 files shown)


Sources/StreamVideo/WebRTC/PeerConnectionFactory.swift
```diff
@@ -20,7 +20,7 @@ final class PeerConnectionFactory: @unchecked Sendable {
         )
         let decoderFactory = RTCDefaultVideoDecoderFactory()
         return RTCPeerConnectionFactory(
-            audioDeviceModuleType: .platformDefault,
+            audioDeviceModuleType: .audioEngine,
```
martinmitrevski, line 23: why do we change this here?

ipavlidakis: The platformDefault ADM causes the prompts to appear while the audioEngine based one doesn't. We are already planning on migrating to the AudioEngine ADM so we can also do it here and resolve the issue.

martinmitrevski (approved):

PR #1004, opened 2025-12-01
[Enhancement]Handle video moderation event

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1226/enhancementimplement-blur-and-warning-moderation-events

### Docs

https://github.com/GetStream/docs-content/pull/824

### 🎯 Goal

Respond on moderation events and apply relative actions.

### 🛠 Implementation

- The revision adds 2 new VideoFilters (blur and pixelate).
- ⚠️ Breaking change: the revision removes the ModerationBlurViewModifier.swift as now the logic is being applied on the published stream rather than on the received.

### 🧪 Manual Testing Notes

- ModerationVideoPolicies are configurable via the debug menu. Access to the moderation dev app is required.

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/Sources/Components/AppEnvironment.swift
+++ b/DemoApp/Sources/Components/AppEnvironment.swift
@@ -646,12 +646,12 @@ extension AppEnvironment {
             }
         }
 
-        var value: ModerationManager.VideoPolicy {
+        var value: Moderation.VideoPolicy {
             switch self {
             case .blur(let duration):
-                return ModerationManager.VideoPolicy(duration: duration, videoFilter: .blur)
+                return Moderation.VideoPolicy(duration: duration, videoFilter: .blur)
             case .pixelate(let duration):
-                return ModerationManager.VideoPolicy(duration: duration, videoFilter: .pixelate)
+                return Moderation.VideoPolicy(duration: duration, videoFilter: .pixelate)
             }
         }
     }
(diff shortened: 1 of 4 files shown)


DemoApp/Sources/Views/Login/DebugMenu.swift
```diff
@@ -499,6 +528,19 @@ struct DebugMenu: View {
         }
     }
 
+    @ViewBuilder
+    private var customModerationVideoView: some View {
+        Button {
+            presentsCustomModerationVideoPolicyDuration = true
+        } label: {
+            Label {
```
martinmitrevski: Isn't a Text here enough instead?

Sources/StreamVideo/Call.swift
```diff
@@ -527,14 +532,15 @@ public class Call: @unchecked Sendable, WSEventsSubscriber {
         await callController.updateTrackSize(trackSize, for: participant)
     }
 
-    /// Sets a `videoFilter` for the current call.
-    /// - Parameter videoFilter: A `VideoFilter` instance representing the video filter to set.
+    /// Sets a `VideoFilter` for the current call.
+    /// - Parameter videoFilter: Desired filter; pass `nil` to clear it.
     public func setVideoFilter(_ videoFilter: VideoFilter?) {
+        moderation.setVideoFilter(videoFilter)
```
martinmitrevski, line 538: why do we need it here as well? Imagine you put some filter (e.g. image bg), and then the moderation filter kicks in - it would be the same one, right?

ipavlidakis: we are reusing the same videoFilter pipeline we have. Once moderation kicks we replace any existing videoFilter (if any) with the moderation filter. Once moderation blur stops we reinstate any videoFilters that were selected prior to the moderation. VideoFilters are quite expensive and having to apply more than one at every frames is inefficient.

DemoApp/Sources/Views/Login/DebugMenu.swift
ipavlidakis: That's true. If there is another commit i'll change it otherwise we can do it on the first chance we get back here.

PR #1028, opened 2026-01-15
[Swift 6 Migration]StreamVideo

ipavlidakis:
### 🛠 Implementation

This revision migrates StreamVideo and StreamVideoTests to Swift 6 with Complete concurrency

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideo/Utils/Swift6Migration/Sendable+Extensions.swift
+++ b/Sources/StreamVideo/Utils/Swift6Migration/Sendable+Extensions.swift
@@ -30,6 +30,7 @@ extension AnyPublisher: @retroactive @unchecked Sendable {}
 extension Publishers.Filter: @retroactive @unchecked Sendable {}
 /// Allows audio buffers to cross concurrency boundaries.
 extension AVAudioPCMBuffer: @retroactive @unchecked Sendable {}
+extension CIContext: @retroactive @unchecked Sendable {}
 #else
 extension AnyCancellable: @unchecked Sendable {}
 extension AVCaptureDevice: @unchecked Sendable {}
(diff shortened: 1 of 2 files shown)


Sources/StreamVideo/Utils/SendableBox/SendableBox.swift
```diff
@@ -0,0 +1,54 @@
+//
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Namespace for wrapping non-sendable closures in tests.
+enum SendableBox {}
```
martinmitrevski, line 8: Why do we need these?

Sources/StreamVideo/Utils/SystemEnvironment.swift
```diff
@@ -39,9 +39,9 @@ enum SystemEnvironment {
 
     static var isTests: Bool {
         #if STREAM_TESTS
-        return NSClassFromString("XCTest") != nil
+        return true
         #else
-        return false
+        return NSClassFromString("XCTest") != nil
```
martinmitrevski, line 42: why do we change this? It's safer to just return false here.

Sources/StreamVideo/Utils/SendableBox/SendableBox.swift
ipavlidakis: We need it to wrap the generic Combine.Publisher type, which is a protocol and we cannot simply extend it. We are also using it to wrap a closure that was causing a warning again because the closure couldn't be marked as Sendable.

Sources/StreamVideo/Utils/SystemEnvironment.swift
ipavlidakis: good catch! i changed that for some testing and need to revert it

PR #1059, opened 2026-02-17
[Fix]Join returns only when SFU is connected

ipavlidakis:
### 🔗 Issue Links

Resolves:
- https://linear.app/stream/issue/IOS-1393/video-speaker-value-is-rewritten-when-using-join-and-ring
- https://linear.app/stream/issue/IOS-1422/bugjoincall-should-return-after-sfu-connection-is-established

### 🎯 Goal

Align `Call.join` completion with the SFU connection lifecycle so a successful join now completes only after the connected handshake reaches the expected stage, while preserving local call settings and improving coverage around this behavior.

### 🛠 Implementation

- Updated join completion flow to rely on `PassthroughSubject<JoinCallResponse, Error>` callbacks in coordinator/controller paths.
- Preserved the initial `JoinCallResponse` until the SFU handshake transitions to connected for deterministic completion timing.
- Added completion failure propagation on error-stage transitions.

### 🧪 Manual Testing Notes

Full regression cycle is necessary. Flows to test:
- Join call
- Ring and receive calls
- JoinAndRing

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Join flow now aligns with SFU handshake and reports failures earlier during accept.
  * Speaker remains disabled while ringing in ring-and-join flows.

* **Improvements**
  * Join completion now waits for audio session readiness for more reliable connections.
  * Audio activation supports deferred activation for more consistent behavior.
  * Picture-in-Picture lifecycle and stability improved.
  * Call controls respect call settings more consistently.
  * Join errors surface as a standardized "Operation timed out" error.

* **New Features**
  * Added join policy options and a fast-reconnect path to improve connection behavior.

### DIFF
--- a/StreamVideoTests/IntegrationTests/Call_IntegrationTests.swift
+++ b/StreamVideoTests/IntegrationTests/Call_IntegrationTests.swift
@@ -381,13 +381,15 @@ final class Call_IntegrationTests: XCTestCase, @unchecked Sendable {
             group.addTask {
                 try await creatorFlow
                     .perform { try await $0.call.join() }
-                    .assertEventuallyInMainActor { $0.call.state.participants.endIndex == 2 }
+                    .subscribe(for: CustomVideoEvent.self)
+                    .assertEventually { (event: CustomVideoEvent) in event.custom["state"] == "joined" }
                     .perform { try await $0.call.end() }
             }
 
             group.addTask {
                 try await participantUserFlow
                     .perform { try await $0.call.join() }
+                    .perform { try await $0.call.sendCustomEvent(["state": "joined"]) }
                     .assertEventuallyInMainActor { $0.call.streamVideo.state.activeCall == nil }
             }
 
@@ -608,8 +610,8 @@ final class Call_IntegrationTests: XCTestCase, @unchecked Sendable {
     func test_join_livestream_whenCallIsInBackstageOnlyHostCanJoin_thenAfterCallGoesLiveAnyOtherParticipantCanJoin() async throws {
         let callId = String.unique
         let participant = String.unique
-        let joinAheadTimeSeconds: Double = 5
-        let startingDate = Date(timeIntervalSinceNow: 10)
+        let joinAheadTimeSeconds: Double = 10
+        let startingDate = Date(timeIntervalSinceNow: joinAheadTimeSeconds * 2)
         let joiningDate = Date(timeIntervalSinceNow: joinAheadTimeSeconds + 2)
 
         try await helpers
@@ -764,7 +766,8 @@ final class Call_IntegrationTests: XCTestCase, @unchecked Sendable {
         try await withThrowingTaskGroup(of: Void.self) { group in
             group.addTask {
                 try await user1CallFlowAfterCallCreation
-                    .assertEventuallyInMainActor { $0.call.state.participants.endIndex == 2 }
+                    .subscribe(for: CustomVideoEvent.self)
+                    .assertEventually { (event: CustomVideoEvent) in event.custom["state"] == "joined" }
                     .perform { try await $0.call.pinForEveryone(userId: user2, sessionId: user2SessionId) }
                     .assertEventuallyInMainActor { $0.call.state.participantsMap[user2SessionId]?.pin != nil }
             }
@@ -772,6 +775,7 @@ final class Call_IntegrationTests: XCTestCase, @unchecked Sendable {
             group.addTask {
                 try await user2CallFlow
                     .perform { try await $0.call.join() }
+                    .perform { try await $0.call.sendCustomEvent(["state": "joined"]) }
                     .assertEventuallyInMainActor { $0.call.state.participantsMap[user2SessionId]?.pin != nil }
             }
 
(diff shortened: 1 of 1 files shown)


Sources/StreamVideo/Controllers/CallController.swift
```diff
@@ -57,7 +57,7 @@ class CallController: @unchecked Sendable {
     private var cachedLocation: String?
     private let initialCallSettings: CallSettings
 
-    private var joinCallResponseSubject = CurrentValueSubject<JoinCallResponse?, Error>(nil)
+    private var joinCallResponseSubject = PassthroughSubject<JoinCallResponse, Error>()
     private var joinCallResponseFetchObserver: AnyCancellable?
```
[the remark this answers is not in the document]
martinmitrevski: @ipavlidakis does this make any sense?

Sources/StreamVideo/Utils/AudioSession/CallAudioSession.swift
```diff
@@ -12,6 +12,8 @@ final class CallAudioSession: @unchecked Sendable {
 
     @Injected(\.audioStore) private var audioStore
 
+    private enum DisposableKey: String { case deferredActivation }
```
martinmitrevski, line 15: is this needed just for a string? Maybe we just define a string value instead?

martinmitrevski (approved):
looks good to me, we should test this a bit before releasing

Sources/StreamVideo/Utils/AudioSession/CallAudioSession.swift
ipavlidakis: I added it as I have that pattern in other places. Happy to change it to a simple string for now

Sources/StreamVideo/WebRTC/v2/PeerConnection/RTCPeerConnectionCoordinator.swift
```diff
@@ -83,6 +83,9 @@ class RTCPeerConnectionCoordinator: @unchecked Sendable {
             .eraseToAnyPublisher()
     }
 
+    private let connectionStateSubject: CurrentValueSubject<RTCPeerConnectionState, Never> = .init(.new)
+    var connectionStatePublisher: AnyPublisher<RTCPeerConnectionState, Never> { connectionStateSubject.eraseToAnyPublisher() }
```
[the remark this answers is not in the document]
ipavlidakis: The Coordinator is always initiated with a new PeerConnection. So the scenario causing concerns is not reachable.

Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Joining.swift
```diff
@@ -411,13 +421,46 @@ extension WebRTCCoordinator.StateMachine.Stage {
...
+        /// The function waits in parallel for:
+        /// - `audioStore` to report `isActive == true`
+        /// - `audioStore` to report a non-empty `currentRoute`
+        /// within the provided `timeout`.
+        ///
+        /// - Parameter timeout: Maximum number of seconds to wait for both
+        ///   conditions before failing.
+        /// - Throws: If either readiness condition does not arrive before
+        ///   `timeout`.
+        private func ensureAudioSessionIsReady() async throws {
```
martinmitrevski, line 444: I think we discussed this once, so the audio session is always active even if you don't have send-audio permission? Same for the `currentRoute`.

Sources/StreamVideo/Call.swift
```diff
@@ -608,8 +616,30 @@ public class Call: @unchecked Sendable, WSEventsSubscriber {
...
+        /// Upon `Call.leave` we remove the call from the cache. Any further actions that are required
+        /// to happen on the call object (e.g. rejoin) will need to fetch a new instance from `StreamVideo`
+        /// client.
+        callCache.remove(for: cId)
+        outgoingRingingController = nil
+
+        // Reset the activeAudioFilter
+        setAudioFilter(nil)
+
+        let strongSelf = self
```
martinmitrevski, line 632: why's this needed?

Sources/StreamVideoSwiftUI/CallViewModel.swift
```diff
@@ -589,7 +589,8 @@ open class CallViewModel: ObservableObject {
                     callType: callType,
                     callId: callId,
                     members: [],
-                    customData: customData
+                    customData: customData,
+                    policy: .peerConnectionReadinessAware(timeout: 2)
```
martinmitrevski, line 593: do we just join the call after the timeout, or do we throw an error?

martinmitrevski (approved):

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: We just join, we don't throw

Sources/StreamVideo/Call.swift
ipavlidakis: Because once we nullify the ringing and activeCall, there is nothing else holding a reference to the call, making the postNotification using a nil value.

Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Joining.swift
ipavlidakis: Yes the audioSession is configured and activated regardless of the audio send capabillity, as it's needed for playback. We have an audioSession policy that configures the audioSession differently based on the send-audio capability but that simply changes the AVAudioSession.Mode to `playback` or `playAndRecord`.

PR #1108, opened 2026-04-09
[Enhancement]Add call join interception hook before final call entry

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1607/feature-requestincall-status-sync-between-participants

### 🎯 Goal

Add a join interception hook that lets apps delay or abort call entry after the join response has been applied locally, so participant readiness/status can be synchronized before the user is treated as fully in-call.

### 📝 Summary

- Added `CallJoinIntercepting` and `CallJoinInterceptionError` to support app-defined join interception.
- Extended `Call.join(...)` with an optional `joinInterceptor` parameter.
- Updated the call joining stage to run the interceptor after local join-state update and before setting `activeCall`.
- Added a bounded timeout path for interception using a new `withFirstTaskCompleted` concurrency helper.
- Ensured interception failures trace the error and leave the call cleanly with `join.interception.failed`.
- Exposed `callJoinInterceptor` on `CallViewModel` and forwarded it through both `enterCall` and `joinAndRingCall`.
- Added a demo-only `DemoCallJoinInterceptor` for ringing flows using a custom `participant.ready` event.
- Added unit coverage for interception success, failure, timeout, non-cooperative interception, and the new concurrency helper.
- Updated changelog for the new client-facing join interception API.

### 🛠 Implementation

The SDK now supports an optional interception step during `Call.join(...)`. The interceptor runs after the backend join request succeeds and the local `CallState` has been updated, but before the call is promoted to `StreamVideo.State.activeCall`.

To keep this step bounded, the join flow now races the interceptor against a timeout using `withFirstTaskCompleted`. This is intentionally implemented with unstructured tasks so the join flow can stop waiting even if the interceptor is suspended in a cancellation-insensitive operation (for example, a continuation-backed callback bridge). If the timeout wins, the join continues. If the interceptor throws, the flow surfaces a `CallJoinInterceptionError`, traces the failure, and leaves the call so the SDK does not remain joined underneath a failed UI flow.

On the SwiftUI side, `CallViewModel` exposes `callJoinInterceptor` and forwards it through both normal join/start paths and `joinAndRingCall`, so the feature is applied consistently. The demo app includes a sample interceptor that waits for another participant’s readiness signal via a custom event, but it is explicitly guarded to act only for ringing-call flows.

### 🧪 Manual Testing Notes

- Start a ringing call between two demo users and verify the local participant does not fully enter the call until the remote readiness event is observed.
- Verify that the demo interceptor does not affect non-ringing demo join flows.
- Force the interceptor to exceed the configured timeout and verify the join still completes.
- Force the interceptor to throw and verify the join fails cleanly without leaving an active/joined call behind.
- Verify both normal join/start flows and `joinAndRingCall` honor `callJoinInterceptor`.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Added optional join interception hook to delay or abort call entry
  * New CallJoinIntercepting protocol and per-call CallViewModel.callJoinInterceptor
  * Join interception timeout defaults (prod: 5s, test: 10s)
  * New call-level tracing API to submit WebRTC traces
  * Demo app includes a sample join interceptor

* **Documentation**
  * CHANGELOG updated with join interception details

### DIFF
--- a/DemoApp/Sources/Views/CallView/DemoCallContainerView.swift
+++ b/DemoApp/Sources/Views/CallView/DemoCallContainerView.swift
@@ -21,6 +21,7 @@ internal struct DemoCallContainerView: View {
         let callViewModel = CallViewModel()
         callViewModel.participantAutoLeavePolicy = AppEnvironment.autoLeavePolicy.policy
         callViewModel.isPictureInPictureEnabled = AppEnvironment.pictureInPictureIntegration == .enabled
+        callViewModel.callJoinInterceptor = DemoCallJoinInterceptor()
         _viewModel = StateObject(wrappedValue: callViewModel)
         _chatViewModel = StateObject(wrappedValue: .init(callViewModel))
         self.callId = callId
--- a/DemoApp/Sources/Views/CallView/DemoCallContainerView.swift
+++ b/DemoApp/Sources/Views/CallView/DemoCallContainerView.swift
@@ -21,7 +21,7 @@ internal struct DemoCallContainerView: View {
         let callViewModel = CallViewModel()
         callViewModel.participantAutoLeavePolicy = AppEnvironment.autoLeavePolicy.policy
         callViewModel.isPictureInPictureEnabled = AppEnvironment.pictureInPictureIntegration == .enabled
-        callViewModel.callJoinInterceptor = DemoCallJoinInterceptor()
+        callViewModel.callJoinInterceptor = AppEnvironment.callJoinInterceptor.value
         _viewModel = StateObject(wrappedValue: callViewModel)
         _chatViewModel = StateObject(wrappedValue: .init(callViewModel))
         self.callId = callId
--- a/Sources/StreamVideo/Utils/Extensions/Concurrency/Task+FirstTaskCompleted.swift
+++ b/Sources/StreamVideo/Utils/Extensions/Concurrency/Task+FirstTaskCompleted.swift
@@ -127,8 +127,12 @@ private final class FirstTaskCompletedCoordinator<First: Sendable, Second: Senda
 
     func setFirstTask(_ task: Task<Void, Never>) {
         let shouldCancel = queue.sync { () -> Bool in
+            guard resolvedResult == nil else {
+                return true
+            }
+
             firstTask = task
-            return resolvedResult != nil
+            return false
         }
 
         if shouldCancel {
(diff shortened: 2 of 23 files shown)


Sources/StreamVideo/Utils/Extensions/Concurrency/Task+FirstTaskCompleted.swift
```diff
@@ -0,0 +1,182 @@
...
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Indicates which of two concurrently started tasks completed first.
+///
+/// Completion means the task either returned a value or threw an error.
+///
+/// - Important: The losing task is cancelled but not awaited. If it does not
```
martinmitrevski, line 13: Do customers need to do something around the cancellation? If yes, it might be tricky / error prone.

Sources/StreamVideo/Utils/Extensions/Concurrency/Task+FirstTaskCompleted.swift
```diff
@@ -0,0 +1,182 @@
...
+            secondTask = task
+            return false
+        }
+
+        if shouldCancel {
+            task.cancel()
+        }
+    }
+
+    func cancelFirstTask() {
```
martinmitrevski: maybe we just pass the task as a param and have just one method? (Both are the same atm)

Sources/StreamVideo/Utils/Extensions/Concurrency/Task+FirstTaskCompleted.swift
```diff
@@ -0,0 +1,182 @@
...
+
+        guard resolution.0 else {
+            return false
+        }
+
+        resolution.1?.resume(returning: result)
+        return true
+    }
+
+    func setFirstTask(_ task: Task<Void, Never>) {
```
martinmitrevski: we can also change this to one method

DemoApp/Sources/Views/CallView/DemoCallContainerView.swift
```diff
@@ -21,6 +21,7 @@ internal struct DemoCallContainerView: View {
         let callViewModel = CallViewModel()
         callViewModel.participantAutoLeavePolicy = AppEnvironment.autoLeavePolicy.policy
         callViewModel.isPictureInPictureEnabled = AppEnvironment.pictureInPictureIntegration == .enabled
+        callViewModel.callJoinInterceptor = DemoCallJoinInterceptor()
```
martinmitrevski: don't think we should have this by default, just having the example in the repo is fine

martinmitrevski (commented):
Looks great! We should test a bit, and also left few comments - mostly around reducing repetitive code.

Sources/StreamVideo/Utils/Extensions/Concurrency/Task+FirstTaskCompleted.swift
ipavlidakis: In the current way we implemented it (without the use of TaskGroup) they don't.

martinmitrevski (approved):

PR #1109, opened 2026-04-09
[Fix]Premature CallViewModel.callingState transitions from participan…

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1614/enhancementcallviewmodelcallingstate-is-driven-only-by-events-and

### 🎯 Goal

Prevent `CallViewModel.callingState` from moving forward before the call is actually ready, especially when participant updates arrive while the app is backgrounded or while a CallKit-accepted join is still restoring.

### 📝 Summary

- Removed participant-driven `callingState` transitions from `CallViewModel`
- Kept participant updates purely as participant data updates
- Added an explicit CallKit `.joined` -> `.inCall` handoff so CallKit-accepted joins do not remain stuck in `.joining`
- Added unit coverage for CallKit join ordering, reconnect ordering, and outgoing ringing ordering

### 🛠 Implementation

`CallViewModel` previously updated `callingState` from `callParticipants.didSet`, which meant background participant refreshes could push the UI into `.inCall` before the join flow had actually completed.

This PR removes that participant-driven fallback and keeps state progression tied to readiness-related signals instead. While doing that, it also fixes the CallKit restore path: if the active call is restored while the temporary CallKit bridge is still in `.joining`, the later `.joined` event now explicitly hands control back to the regular `setActiveCall` flow so the UI transitions to `.inCall` at the right time.

The test coverage added in `CallViewModel_Tests` verifies:
- participant updates do not advance `.joining` during the CallKit flow
- a later CallKit `.joined` event transitions the view model to `.inCall`
- participant churn does not collapse `.reconnecting` or `.outgoing` early
- the real acceptance / connection restoration path still transitions to `.inCall`

### 🧪 Manual Testing Notes

- Accept a ringing call while the app is in background and screen is unlocked
- Accept from the CallKit notification / system UI and verify the UI goes to `.joining` first and then `.inCall`
- Repeat with a cold-start accept from CallKit
- Verify participant updates while backgrounded do not surface the in-call UI before the call is actually ready
- Verify ringing / reconnect flows still end in `.inCall` once readiness is reached

### ☑️ Contributor Checklist

## Bug Fixes
* Fixed calling state management to prevent premature transitions to "in call" when participants update before the call initialization is complete.

### DIFF
--- a/Sources/StreamVideoSwiftUI/CallViewModel.swift
+++ b/Sources/StreamVideoSwiftUI/CallViewModel.swift
@@ -147,11 +147,7 @@ open class CallViewModel: ObservableObject {
     @Published public var outgoingCallMembers = [Member]()
 
     /// Dictionary of the call participants.
-    @Published public private(set) var callParticipants = [String: CallParticipant]() {
-        didSet {
-            updateCallStateIfNeeded()
-        }
-    }
+    @Published public private(set) var callParticipants = [String: CallParticipant]()
 
     /// Contains info about a participant event. It's reset to nil after 2 seconds.
     @Published public var participantEvent: ParticipantEvent?
@@ -297,17 +293,22 @@ open class CallViewModel: ObservableObject {
         callKitServiceObserver
             .publisher
             .receive(on: DispatchQueue.main)
-            .compactMap {
-                switch $0 {
+            .sink { [weak self] event in
+                guard let self else { return }
+
+                switch event {
                 case let .joining(call):
-                    return call
+                    setCallingState(.joining)
+                    self.call = call
+                case .joined:
+                    guard let call else { return }
+                    // CallKit can report the active call while it still marks
+                    // the bridge as `.joining`. Once `.joined` arrives we need
+                    // an explicit handoff to let the regular in-call UI take over.
+                    setActiveCall(call)
                 default:
-                    return nil
+                    break
                 }
-            }
-            .sink { [weak self] in
-                self?.setCallingState(.joining)
-                self?.call = $0
             }
             .store(in: disposableBag)
     }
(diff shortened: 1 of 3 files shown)


Sources/StreamVideoSwiftUI/CallViewModel.swift
```diff
@@ -284,18 +280,23 @@ open class CallViewModel: ObservableObject {
         callKitServiceObserver
             .publisher
             .receive(on: DispatchQueue.main)
-            .compactMap {
-                switch $0 {
+            .sink { [weak self] event in
```
martinmitrevski, line 296: I might be missing something, but how's the state updated if we don't use call kit?

ipavlidakis: If CallKit is disabled then the CallViewModel is the only managing state internally in various places:

joining: startCall, join, acceptCall
outgoing: startCall, joinAndRing
incoming: subscribeToCallEvents(based on ws events)
reconnecting: (same as now) based on Call reconnectionStatus
inCall: based on Call reconnectionStatus(for all flows including joinAndRing), setActiveCall(for non joinAndRingFlows).

martinmitrevski: yes, but how, now that we delete updateCallStateIfNeeded?

ipavlidakis: `updateCallStateIfNeeded` was only used when the participants were updated. We aren't using anywhere else and the call state shouldn't relying on participants updates

martinmitrevski (approved):

PR #1115, opened 2026-04-16
[Fix]CallKit ringing edge cases for already-handled incoming calls

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1628/bughandle-edge-cases-for-callkit-ringing

### 🎯 Goal

Ensure CallKit stops ringing cleanly when an incoming call was already
handled elsewhere, and preserve the explicit backend leave reason for those
cases.

### 📝 Summary

- detect already-handled incoming calls when the current user already
  accepted, rejected, or missed the call
- detect calls that already ended elsewhere or were rejected by the other
  participants before the local device answers
- propagate explicit leave reasons through the CallKit end-call flow so the
  backend receives the correct rejection reason

### 🛠 Implementation

`CallKitService.checkIfCallWasHandled` now returns an optional leave reason
instead of a boolean, so the incoming-call flow can distinguish why the call
should stop ringing.

When `reportIncomingCall` fetches a call state that has already been handled,
it immediately ends the CallKit call and forwards that explicit leave reason.
`CallKitService.callEnded` now stores the leave reason on the pending
`CallEntry`, and `provider(_:perform:)` for `CXEndCallAction` prefers that
stored reason before falling back to `rejectionReasonProvider`.

### 🧪 Manual Testing Notes

1. Have another user call your device.
2. Immediately after the outgoing ringing starts, have the caller hang up.
3. Verify the CallKit notification appears on your device and disappears immediately.
4. Verify the call does not remain stuck in ringing state in the app.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved CallKit ringing edge cases: incoming calls already handled elsewhere are now ended reliably and rejection reasons are preserved and forwarded so call outcomes (answered, rejected, missed, ended, auto-leave) are reported consistently.
* **Tests**
  * Expanded and improved test coverage for incoming-call handling, leave/reject reasons, missed/rejected scenarios, and auto-leave behavior.


Sources/StreamVideo/CallKit/CallKitService.swift
```diff
@@ -639,32 +658,110 @@ open class CallKitService: NSObject, CXProviderDelegate, @unchecked Sendable {
...
+    ///    another device.
+    /// 3. The creator rejected it.
+    /// 4. Every remaining participant except the current user and creator
+    ///    rejected it.
+    ///
+    /// - Parameter callState: The latest call state fetched from the backend
+    ///   before starting the local ringing timer.
+    /// - Returns: A backend leave reason when the call should stop ringing
+    ///   immediately, or `nil` when the local device should keep ringing.
+    ///   The current mapping is:
```
martinmitrevski: is this mapping aligned with other SDKs?

ipavlidakis: Now they are :) Thanks

martinmitrevski (approved):

PR #1118, opened 2026-04-16
Fix flaky StreamVideo tests

ipavlidakis:
## Summary
- replace brittle sleep-based test waits with state-driven assertions in `Call_IntegrationTests`
- remove a ring-event subscription race from the accept integration flow
- make device deletion checks eventually consistent instead of asserting a single immediate snapshot
- harden state-machine tests that expected no transition so unexpected transitions fail fast
- switch the shared XCTest wait helper to `Task.sleep`
- stabilize timer resubscription coverage and drop duplicate duration coverage from `Call_Tests`

## Testing
- Ran `xcodebuild -project StreamVideo.xcodeproj -scheme StreamVideo -testPlan StreamVideo -destination 'platform=iOS Simulator,name=iPhone 17 Pro,OS=26.4' test`
- Verified targeted integration cases individually with plain `xcodebuild`
- Final full-suite run completed with no failing tests

## Summary by CodeRabbit

* **Tests**
  * Replaced brittle fixed delays with event-based eventual assertions to reduce flakiness.
  * Improved async wait helpers and centralized assertion helpers for more reliable timing and clearer failure reporting.
  * Hardened expectation behavior to tolerate multiple fulfillments and reduced spurious failures.
  * Removed one flaky duration test.


StreamVideoTests/Call/Call_Tests.swift
```diff
@@ -370,42 +370,6 @@ final class Call_Tests: StreamVideoTestCase, @unchecked Sendable {
         )
     }
 
-    // MARK: - Duration
```
ipavlidakis, line 373: That same behavior is already covered directly in CallState_Tests.swift (line 224), CallState_Tests.swift (line 237), and CallState_Tests.swift (line 278).

martinmitrevski (approved):

PR #1119, opened 2026-04-16
[Enhancement]Limit WebRTC rejoin bursts after repeated disconnects

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1626/investigationmultiple-joinrequest

### 🎯 Goal

Stop pathological internal WebRTC rejoin loops from issuing repeated backend `joinCall` requests after unstable or abnormal disconnects, while still allowing isolated valid rejoins over the lifetime of a call.

### 📝 Summary

- Add a rolling rejoin-attempt limit to the internal WebRTC reconnect path
- Enforce the limit only when the reconnection strategy resolves to `.rejoin`
- Leave the call cleanly once the rejoin burst limit is exceeded instead of continuing the loop
- Add production documentation explaining that this is a rolling-window burst guard, not a lifetime cap
- Add focused state-machine tests for rejoin attempt tracking, pruning, limit enforcement, and cleanup reset

### 🛠 Implementation

This change introduces a rolling rejoin window in `WebRTCCoordinator.StateMachine.Stage.Context` with a default of 5 rejoin attempts within 120 seconds.

The enforcement point is in `WebRTCCoordinator+Disconnected.swift`. When the reconnect flow resolves to `.rejoin` directly, or when a fast reconnect is no longer possible and the SDK falls back to `.rejoin`, the coordinator now calls `registerRejoinAttempt()` before transitioning into `RejoiningStage`.

If the attempt is still within budget, the existing rejoin flow continues unchanged. If the recent rejoin budget is exhausted, the SDK stops the internal loop and transitions to `LeavingStage` with reason `rejoin_attempt_limit_exceeded`. This prevents repeated backend `joinCall` bursts without changing the top-level `Call.join()` retry behavior and without changing successful fast reconnect behavior.

The production comments were expanded to make the semantics explicit:
- this is not a lifetime cap on rejoins for the whole call
- old attempts are pruned before each decision
- rejoins spaced far apart, for example 6 minutes apart, remain valid because they fall outside the 120-second window

Test coverage was added at two levels:
- low-level `WebRTCCoordinatorStateMachine_DisconnectedStageTests` verify registration, pruning, and limit enforcement
- higher-level `Call_JoinRecovery_Tests` verify that real `Call.join()` flows still issue legitimate follow-up backend joins, and that the repeated subscriber-disconnect scenario now stops after the configured cap

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Added guarded rejoin logic that enforces a rolling limit on rejoin attempts during recovery.

* **Bug Fixes**
  * Prevents uncontrolled rejoin loops by capping attempts within a time window; excess attempts log a warning and end the reconnection flow.

* **Tests**
  * Added and extended tests verifying rejoin-rate limiting, fast/normal reconnect behavior, and backend join request patterns.

### DIFF
--- a/Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Disconnected.swift
+++ b/Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Disconnected.swift
@@ -156,11 +156,11 @@ extension WebRTCCoordinator.StateMachine.Stage {
                         try transition?(.fastReconnecting(context))
                     } else {
                         context.fastReconnectionAttempts = 0
-                        try transition?(.rejoining(context))
+                        try transitionToRejoiningIfPossible()
                     }
                 case .rejoin:
                     context.fastReconnectionAttempts = 0
-                    try transition?(.rejoining(context))
+                    try transitionToRejoiningIfPossible()
                 case .migrate:
                     context.fastReconnectionAttempts = 0
                     try transition?(.migrating(context))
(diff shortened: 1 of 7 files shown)


Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Disconnected.swift
```diff
@@ -183,6 +183,44 @@ extension WebRTCCoordinator.StateMachine.Stage {
...
+        ///
+        /// Important:
+        /// - This does not cap the total number of rejoins for the whole call.
+        /// - It only caps how many rejoins may happen inside the recent rolling
+        ///   window tracked by `Context.registerRejoinAttempt`.
+        /// - When attempts are spaced far enough apart, earlier attempts age out
+        ///   of the window and new rejoins are allowed.
+        ///
+        /// With the default configuration, a call that rejoins once every
+        /// 6 minutes will continue to rejoin successfully, because each prior
```
martinmitrevski, line 197: isn't this too strict, especially for slower networks?

Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Stage.swift
```diff
@@ -47,6 +47,34 @@ extension WebRTCCoordinator.StateMachine {
...
+            /// disconnects, while still allowing long-running calls to recover
+            /// from occasional isolated rejoin events.
+            ///
+            /// Example:
+            /// - If 5 rejoins happen within 2 minutes, the 6th rejoin inside
+            ///   that same 2-minute window is rejected.
+            /// - If rejoins are 6 minutes apart, older attempts have already
+            ///   aged out of the rolling window before the next one happens, so
+            ///   each rejoin is treated as an isolated recovery and is allowed.
+            var rejoinMaxAttempts: Int = 5
```
martinmitrevski: maybe we allow 10 instead?

martinmitrevski (commented):
Looks good, although I would consider extending the attempts a bit

Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Stage.swift
ipavlidakis: I made the change. Thanks :)

martinmitrevski (approved):

PR #1126, opened 2026-04-20
Iliaspavlidakis/ios 1613 investigationtimeout issues second attempt

ipavlidakis:
### 🔗 Issue Links

Resolve https://linear.app/stream/issue/IOS-1613/investigationtimeout-issues

### 🎯 Goal

Prevent cancelled join attempts from lingering until the join timeout expires, which could surface stale `TimeOutError`s even when a later retry succeeds.

### 📝 Summary

- Propagate cancellation immediately through the Combine-to-async publisher bridge used by call join flows.
- Ensure timed publisher waits cancel their underlying timeout task instead of waiting for the full deadline.
- Add direct cancellation regressions for `firstValue(...)` and `nextValue(...)`.
- Add an integration regression covering: participant joins, cancels a join attempt in progress, retries, and successfully rejoins without later timeout fallout.
- Add docC comments to the changed public async publisher helpers.

### 🛠 Implementation

The fix keeps the join timeout behavior intact for genuine timeout cases, but changes cancellation semantics so a cancelled caller exits promptly with `CancellationError`.

In the async publisher bridge:
- `firstValue()` now treats cancellation during stream consumption as cancellation instead of a generic “Task produced no value” client error.
- `firstValue(timeoutInSeconds:)` wraps the timeout task in a cancellation handler so cancelling the caller also cancels the underlying timed wait immediately.
- `nextValue(...)` inherits that behavior through the shared helper path.

Test coverage now exists at three levels:
- unit coverage for direct publisher cancellation behavior
- timeout helper cancellation coverage
- an end-to-end integration scenario for cancelling a join in progress and retrying successfully
- 
### 🧪 Manual Testing Notes

- Start a 1:1 call.
- Let participant A join.
- Start participant B join, cancel it mid-join, then retry.
- Confirm both participants join successfully and no delayed timeout error appears afterward.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation handling so cancelled operations stop promptly and report cancellation instead of generic failures.
  * Ensured timeout tasks are cancelled immediately when an awaiting operation is cancelled.
  * Joining now leaves the call on final handoff timeout and stops retrying, preserving call state.

* **Tests**
  * Added and updated tests covering cancellation, timeout, join timeout leave behavior, and join/retry recovery scenarios.

### DIFF
--- a/Sources/StreamVideo/WebRTC/v2/Extensions/Foundation/Task+Timeout.swift
+++ b/Sources/StreamVideo/WebRTC/v2/Extensions/Foundation/Task+Timeout.swift
@@ -16,7 +16,7 @@ extension Task where Failure == any Error {
         line: UInt = #line,
         operation: @Sendable @escaping () async throws -> Success
     ) {
-        self = Task(priority: priority) {
+        self = Task<Success, Failure>(priority: priority) {
             try await withThrowingTaskGroup(of: Success.self) { group in
 
                 /// Add the operation to perform as the first task.
@@ -39,7 +39,7 @@ extension Task where Failure == any Error {
                 /// This is default for task groups to account for when there aren't any pending tasks.
                 /// Awaiting on an empty group immediately returns 'nil' without suspending.
                 guard let result = try await group.next() else {
-                    throw ClientError("Task produced no value", file, line)
+                    throw _Concurrency.CancellationError()
                 }
 
                 /// If we reach this, it means we have a value before the timeout.
--- a/Sources/StreamVideo/WebRTC/v2/Extensions/Foundation/Task+Timeout.swift
+++ b/Sources/StreamVideo/WebRTC/v2/Extensions/Foundation/Task+Timeout.swift
@@ -4,6 +4,8 @@
 
 import Foundation
 
+private typealias SwiftCancellationError = CancellationError
+
 /// An extension to the `Task` type that adds timeout functionality.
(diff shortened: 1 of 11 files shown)


Sources/StreamVideo/WebRTC/v2/Extensions/Foundation/Task+Timeout.swift
```diff
@@ -39,7 +39,7 @@ extension Task where Failure == any Error {
                 /// This is default for task groups to account for when there aren't any pending tasks.
                 /// Awaiting on an empty group immediately returns 'nil' without suspending.
                 guard let result = try await group.next() else {
-                    throw ClientError("Task produced no value", file, line)
+                    throw _Concurrency.CancellationError()
```
martinmitrevski: how about this one?

martinmitrevski (approved):

Sources/StreamVideo/WebRTC/v2/Extensions/Foundation/Task+Timeout.swift
ipavlidakis: Unfortunately, because we were inside a task the `_Concurrency` namespace was inferred. In order to be able to use the correct error, as everywhere else i simply namespaced it to `Swift.CancellationError`.

PR #1134, opened 2026-04-22
[Enhancement]Add startCall high-scale livestream publisher hint

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1623/enhancementimplement-high-tier-sfu-hinting

### 🎯 Goal

Add support for the high-scale livestream publisher join hint across the iOS SDK so clients can explicitly mark publisher joins that should use the backend’s high-scale livestream routing behavior, and expose a Demo app toggle to exercise that end-to-end.

### 📝 Summary

- add `CreateCallOptions.highScaleLivestreamPublisherHint`
- forward the hint into `JoinCallRequest.hintHighScaleLivestreamPublisher`
- persist the hint in `CallController` so recovery joins reuse it
- expose the hint through `CallViewModel.startCall(...)`
- add a Demo app feature flag to enable/disable the hint, defaulting to disabled
- add unit coverage for core SDK propagation, recovery joins, and SwiftUI forwarding

### 🛠 Implementation

On the core SDK side, the high-scale livestream publisher hint is now part of `CreateCallOptions` and is mapped to `JoinCallRequest.hintHighScaleLivestreamPublisher` before the backend `joinCall` request is sent. `CallController` caches the value from the explicit join invocation so recovery joins continue to send the same hint even when reconnection paths no longer carry `CreateCallOptions`.

On the SwiftUI side, `CallViewModel.startCall(...)` now accepts the optional hint and forwards it through the create-and-join path. This keeps the feature explicit at the call site rather than storing it as mutable view model state. `joinAndRingCall(...)` was intentionally left unchanged.

In the Demo app, a new feature flag was added under the existing debug feature-flags menu in a dedicated file. Demo `startCall(...)` call sites now pass the flag explicitly so the end-to-end behavior can be toggled without changing code.

The change is covered with focused tests in the core SDK for request shape and recovery propagation, and in SwiftUI for `startCall(...)` forwarding.

### 🧪 Manual Testing Notes

1. Open the Demo app debug menu.
2. Under `Feature Flags`, toggle `High-Scale Livestream Hint`.
3. Start a call through a Demo `startCall(...)` flow with the flag disabled and confirm the call joins normally.
4. Enable the flag and repeat the same flow.
5. Optionally verify on the backend/request inspection side that the join request includes `hint_high_scale_livestream_publisher = true` when enabled.
6. Confirm reconnect/recovery behavior still works after the initial join.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Call join now supports an optional high-scale livestream publisher hint to enable optimized routing for large-scale livestreams.

* **Debug**
  * New debug toggle to enable/disable the high-scale livestream publisher hint from the app’s debug menu.

* **Documentation**
  * CHANGELOG updated to reflect availability of the high-scale livestream publisher hint.

* **Tests**
  * Added tests verifying the hint is propagated through join and recovery join flows.

### DIFF
--- a/Sources/StreamVideo/Controllers/CallController.swift
+++ b/Sources/StreamVideo/Controllers/CallController.swift
@@ -117,8 +117,6 @@ class CallController: @unchecked Sendable {
     ///   - callSettings: The current call settings.
     ///   - videoOptions: Configuration options about the video.
     ///   - options: Create call options.
-    ///              `CreateCallOptions.highScaleLivestreamPublisherHint` is
-    ///              forwarded only to backend join requests.
     ///   - migratingFrom: If SFU migration is being performed.
     ///   - ring: Whether ringing events should be handled.
     ///   - notify: Whether users should be notified about the call.
--- a/Sources/StreamVideo/Controllers/CallController.swift
+++ b/Sources/StreamVideo/Controllers/CallController.swift
@@ -56,7 +56,6 @@ class CallController: @unchecked Sendable {
     private let videoConfig: VideoConfig
     private let webRTCCoordinatorFactory: WebRTCCoordinatorProviding
     private var cachedLocation: String?
-    private var cachedHighScaleLivestreamPublisherHint: Bool?
     private let initialCallSettings: CallSettings
 
     private var webRTCClientSessionIDObserver: AnyCancellable?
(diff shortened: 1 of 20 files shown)


Sources/StreamVideo/Controllers/CallController.swift
```diff
@@ -116,6 +117,8 @@ class CallController: @unchecked Sendable {
     ///   - callSettings: The current call settings.
     ///   - videoOptions: Configuration options about the video.
     ///   - options: Create call options.
+    ///              `CreateCallOptions.highScaleLivestreamPublisherHint` is
+    ///              forwarded only to backend join requests.
```
martinmitrevski: what does that mean?

martinmitrevski (approved):

Sources/StreamVideo/Controllers/CallController.swift
ipavlidakis: The intention was to mark this as information that is only being forwarded and not used locally but yeah i see the problem. I understand how that is exaggerating (`backend`) and not really needed (as all createOptions are being sent when on requests).

PR #1144, opened 2026-05-18
[CI] Lock protobuf dependency version on CocoaPods

testableapple:
## Test

- https://github.com/GetStream/stream-video-swift/actions/runs/26032917085

## Summary by CodeRabbit

* **Chores**
  * Updated core framework dependency to a stable version for improved consistency and reliability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-video-swift/pull/1144?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/StreamVideo-XCFramework.podspec
+++ b/StreamVideo-XCFramework.podspec
@@ -20,7 +20,7 @@ Pod::Spec.new do |spec|
   spec.vendored_frameworks = "#{spec.module_name}.xcframework", 'Frameworks/StreamWebRTC.xcframework'
   spec.preserve_paths = "#{spec.module_name}.xcframework/*", 'Frameworks/*'
 
-  spec.dependency('SwiftProtobuf', '~> 1.29')
+  spec.dependency('SwiftProtobuf', '1.37.0')
 
   spec.prepare_command = <<-CMD
     mkdir -p Frameworks/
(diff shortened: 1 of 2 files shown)


StreamVideo-XCFramework.podspec
```diff
@@ -20,7 +20,7 @@ Pod::Spec.new do |spec|
   spec.vendored_frameworks = "#{spec.module_name}.xcframework", 'Frameworks/StreamWebRTC.xcframework'
   spec.preserve_paths = "#{spec.module_name}.xcframework/*", 'Frameworks/*'
 
-  spec.dependency('SwiftProtobuf', '~> 1.29')
+  spec.dependency('SwiftProtobuf', '1.37.0')
```
martinmitrevski: shouldn't it be 1.30.0, like on spm?

ipavlidakis: No, it should be pointing to the last version that was working before it starts failing on CI, which was the version before the one released 3 days ago, aka 1.37.0.

We should also pin it in the Package.swift too and the xcode project

martinmitrevski: hmm, but we're using 1.30.0 for a while, isn't it better to keep that one?

ipavlidakis: I see what you mean, Package.swift has it pinned to 1.30.0. 🤔 maybe for consistency we can pin to 1.30.0 (it was a mistake that pods where misaligned with package.swift) and then we can upgrade in a seperate step the version on both package.swift and pods to 1.37.0

martinmitrevski: yes, no need to do it now when we upgrade webrtc

testableapple: updated

ipavlidakis (approved):
I just checked and the xcodeproj is also already pinned to 1.30.0. LGTM ✅

martinmitrevski (approved):
lgtm

PR #1150, opened 2026-05-26
[Fix]Cancel WebRTC stage work before transitions

ipavlidakis:
### 🎯 Goal

We observed a crash that was caused on the AudioIORemote thread and on SendVideoChannel, inside WebRTC. The source of those issues was that after call has been left and its PeerConnections were closed, our flows continued in the background. That was causing audio and video frames to get captured and attempted to sent, which was causing the crash on WebRTC.

This revision now prevents stale WebRTC join and peer-connection readiness tasks from completing after the state machine has already moved to another stage.

### 📝 Summary

- Cancel in-flight `JoiningStage` work when transitioning away.
- Cancel pending `PeerConnectionPreparingStage` readiness waits when transitioning away.
- Preserve WebRTC stage transition tracing for `PeerConnectionPreparingStage`.
- Add DocC comments and focused unit tests for transition-away cleanup behavior.

### 🛠 Implementation

`JoiningStage` and `PeerConnectionPreparingStage` now clear their `DisposableBag` in `willTransitionAway()`, cancelling any stage-owned async work before another stage owns the call lifecycle.

`PeerConnectionPreparingStage` now treats `CancellationError` differently from peer-connection readiness timeouts: timeouts still allow the join to continue, while cancellation exits without reporting join completion or transitioning to `.joined`.

### 🧪 Manual Testing Notes

- Join a call
- While joining leave
- Quickly repeat the steps on the same call
- No crash should be observed
- At some point after joining the call (without leaving) audio should be working as expecte

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### 🐞 Fixed
 - An audio issue that can cause the local AVAudioEngine to fail to initialise.[#1148](https://github.com/GetStream/stream-video-swift/pull/1148)
+- Prevent calls from advancing from stale WebRTC join work after the connection flow has moved on. [#1150](https://github.com/GetStream/stream-video-swift/pull/1150)
 
 # [1.47.0](https://github.com/GetStream/stream-video-swift/releases/tag/1.47.0)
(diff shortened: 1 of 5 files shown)


martinmitrevski (approved):
lgtm ✅

PR #1156, opened 2026-05-29
[Enhancement] Add CallKit handoff for in-app call flows

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1732/blinkcallkit-controls-when-the-call-answered-from-incoming-callscreen

### 🎯 Goal

Allow calls started or accepted from the app UI to be handed over to CallKit so the system UI, audio session ownership, and CallKit controls stay in sync even when the user did not answer from CallKit's native UI.

### 📝 Summary

- Added an internal `CallKitExternalAdapter` that observes SDK `activeCall` and `ringingCall` state and mirrors eligible in-app call flows to CallKit.
- Added CallKit service hooks for externally handled incoming and outgoing calls.
- Added `CXStartCallAction` handling for outgoing calls started from in-app UI.
- Marked in-app CallKit handoffs with `JoinSource.callKit` so WebRTC respects CallKit audio-session ownership.
- Added focused CallKit tests for external incoming/outgoing handoff behavior.

### 🛠 Implementation

`CallKitAdapter` now wires the active `StreamVideo` client into both the existing CallKit service and the new external adapter.

`CallKitExternalAdapter` watches SDK call state transitions:
- outgoing `ringingCall` created by current user triggers `CXStartCallAction`;
- that same call becoming active reports the outgoing CallKit call as connected;
- incoming calls are ignored by this adapter because they already enter CallKit through incoming call reporting.

`CallKitService` now stores CallKit update metadata with each call entry, handles outgoing start actions, and exposes internal APIs for in-app handoff:
- `reportExternalOutgoingCall(_:)`
- `reportExternalConnectedOutgoingCall(_:)`
- `reportExternalIncomingCallConnecting(_:)`

The accepting stage asks CallKit to promote an already reported incoming call before continuing the backend accept flow, keeping CallKit UI aligned with in-app accept.

### 🧪 Manual Testing Notes

- Start an outgoing ringing call from app UI and verify CallKit shows the outgoing call as connecting, then connected after join.
- Receive an incoming CallKit-ringing call, accept from the app UI, and verify CallKit transitions to in-call instead of staying in ringing state.
- End and mute from CallKit controls and verify SDK call state follows existing CallKit paths.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamVideo/CallKit/CallKitService.swift
+++ b/Sources/StreamVideo/CallKit/CallKitService.swift
@@ -26,20 +26,42 @@ open class CallKitService: NSObject, CXProviderDelegate, @unchecked Sendable {
 
     /// Represents a call that is being managed by the service.
     final class CallEntry: Equatable, @unchecked Sendable {
+        /// SDK call object that owns backend and WebRTC lifecycle work.
         var call: Call
+        /// Stable CallKit UUID used for every transaction for this call.
         var callUUID: UUID
+        /// User that created an incoming ringing call, if fetched from backend.
         var createdBy: User?
+        /// Whether CallKit has promoted the call to the active system call.
         var isActive: Bool = false
+        /// Whether the local CallKit ringing timer expired before answer.
         var ringingTimedOut: Bool = false
+        /// Whether backend or another device ended this CallKit entry first.
         var isEndedElsewhere: Bool = false
+        /// Backend rejection or leave reason to apply when CallKit ends it.
         var leaveReason: String?
+        /// Last CallKit presentation metadata associated with this entry.
+        ///
+        /// Incoming calls build this update before calling
+        /// `reportNewIncomingCall`. Outgoing calls build it before requesting
+        /// `CXStartCallAction`. Keeping it with the entry lets follow-up
+        /// CallKit reporting use the same handle, display name, and video
+        /// metadata if a later action needs to refresh the system UI.
+        var callUpdate: CXCallUpdate
 
+        /// Creates a bridge entry between one SDK call and one CallKit UUID.
+        ///
+        /// `callUUID` must stay stable for the whole CallKit lifecycle. CallKit
+        /// treats every action (`start`, `answer`, `mute`, `end`) as belonging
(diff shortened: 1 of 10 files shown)


Sources/StreamVideo/CallStateMachine/Stages/Call+AcceptingStage.swift
```diff
@@ -75,6 +84,14 @@ extension Call.StateMachine.Stage {
                 do {
                     try Task.checkCancellation()
 
+                    // Keep CallKit in sync before the backend accept. This
+                    // covers foreground incoming calls that were already
+                    // reported to CallKit but accepted from the app UI instead
+                    // of the native CallKit answer affordance.
+                    try await callKitService.reportExternalIncomingCallConnecting(call)
```
martinmitrevski, line 91: do we need to always do this? Shouldn't we do it only when going to the background?

Sources/StreamVideo/CallKit/CallKitService.swift
```diff
@@ -497,77 +519,143 @@ open class CallKitService: NSObject, CXProviderDelegate, @unchecked Sendable {
...
+            action.fail()
+        }
+    }
+
     open func provider(
         _ provider: CXProvider,
         perform action: CXAnswerCallAction
     ) {
         guard
-            action.callUUID != active,
```
martinmitrevski, line 505: good if we go through all these changes in the callkit service for me to understand better.

martinmitrevski (commented):
looks good, but we should test a bit. Also, I need some context on the changes in the callkit service.

Sources/StreamVideo/CallStateMachine/Stages/Call+AcceptingStage.swift
ipavlidakis: If we do it only when moving to background i'm afraid we may hit race conditions where the app is being foreground, the callkit report hasn't been completed so no callscreen shows up. 

Reporting always may things simpler.

PR #1161, opened 2026-06-12
[Fix] Call duration starts from local join when a CallJoinInterceptor…

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1758/blinkcall-duration-when-joincallinterceptor-is-involved-its-wrong

### 🥞 Stacked PR 

Depends on #1160

### 🎯 Goal

When a `CallJoinIntercepting` delays call entry, `Call.state.duration` was anchored to the backend session `startedAt`, so users saw elapsed time that included the interceptor's waiting period. The visible duration should start only once the user actually enters the call (reaches the `.joined` state).

### 📝 Summary

- Extracted the duration logic from `CallState` into a new internal `CallDurationTracker` that anchors the duration timer on the backend session start or, when set, on a local start date.
- Added an internal `durationStartOverride` to `CallState` (forwarding to the tracker). The joining stage sets it — only when a join interceptor was involved — right before the call becomes active.
- `CallState.duration` and `CallState.startedAt` remain `@Published` and keep their public semantics; they now republish the tracker's output on the main queue.
- No public API changes. Flows without an interceptor behave exactly as before.

### 🛠 Implementation

- `CallDurationTracker` resolves the timer anchor with explicit precedence (`startOverride ?? session.startedAt ?? session.liveStartedAt`), runs a single timer pipeline, and dedupes repeated session updates via the stored anchor — backend session updates can no longer re-anchor an override. Session end clears the override so a rejoin starts clean.
- `startedAt` keeps reflecting the backend session start even while the duration is anchored locally, preserving its documented meaning.
- The override is set in `JoiningStage.executeJoin` inside the existing MainActor hop that publishes `activeCall`, gated on `input.joinInterceptor != nil` — one condition, surgical.
- `CallState` binds the tracker once in `init`. The binding intentionally uses `receive(on:)` + `sink` (with `MainActor.assumeIsolated`) instead of `sinkTask`: `sinkTask` reuses one task identifier per subscription, so the 1-second timer emissions cancel each other's pending assignments and drop values
- WebRTC reconnects don't re-run the call state machine's joining stage, so the local anchor survives reconnections and the duration doesn't reset.
- CallKit-answered calls are intentionally out of scope and will be handled in a follow-up PR.

### 🧪 Manual Testing Notes

1. In the DemoApp debug menu set **Call Join Interceptor** to **Synchronised**.
2. Ring user B from user A; have B accept in-app while A's interceptor waits for the readiness handshake.
3. Verify B's call duration starts at ~0:00 when the in-call UI appears, not at the time elapsed since the session started.
4. With the interceptor set to **None**, verify durations match the backend session start as before (e.g. joining an ongoing call shows the running call time).

### ☑️ Contributor Checklist

## Release Notes

* **Bug Fixes**
  * Fixed call duration tracking to start from the user's join moment when using join interceptors, rather than from the backend session start time, ensuring accurate duration display.

* **Tests**
  * Enhanced test coverage for call duration tracking behavior and session lifecycle scenarios.

### DIFF
--- a/Sources/StreamVideo/CallState.swift
+++ b/Sources/StreamVideo/CallState.swift
@@ -109,8 +109,9 @@ public class CallState: ObservableObject {
     @Published public internal(set) var isCurrentUserScreensharing: Bool = false
     /// The elapsed duration of the active call session in seconds.
     ///
-    /// The timer begins only after the backend reports a started session and is
-    /// reset when that session ends. Ringing time is not included.
+    /// The timer begins once the backend reports a started session, or from
+    /// the moment the user joined when a join interceptor delayed call entry.
+    /// It is reset when the session ends. Ringing time is not included.
     @Published public internal(set) var duration: TimeInterval = 0
     @Published public internal(set) var statsReport: CallStatsReport?
 
@@ -150,7 +151,20 @@ public class CallState: ObservableObject {
     /// help customize logic, analytics, and UI based on how the call was started.
     var joinSource: JoinSource?
 
-    private var durationCancellable: AnyCancellable?
+    /// When set, `duration` is measured from this local date instead of the
+    /// backend session start.
+    ///
+    /// The join flow sets this once a join interceptor has delayed call entry,
+    /// so the visible duration starts when the user actually joined. It is
+    /// cleared whenever the duration resets (e.g. the session ends).
+    var durationStartOverride: Date? {
+        get { durationTracker.startOverride }
+        set { durationTracker.startOverride = newValue }
+    }
+
+    /// Computes the visible call duration. `duration` and `startedAt`
+    /// republish its output.
+    private let durationTracker = CallDurationTracker()
     private nonisolated let disposableBag = DisposableBag()
 
     /// We mark this one as `nonisolated` to allow us to initialise a state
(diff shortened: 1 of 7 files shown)


Sources/StreamVideo/CallState.swift
```diff
@@ -574,45 +591,36 @@ public class CallState: ObservableObject {
...
-                .assign(to: \.duration, onWeak: self)
+    /// Republishes the tracker's output on the `duration` and `startedAt`
+    /// published properties.
+    ///
+    /// Binding is dispatched once from the nonisolated init onto the main
+    /// actor. The tracker's `CurrentValueSubject`s replay their latest
+    /// value on subscription, so emissions that happen before the bind
+    /// lands are not lost.
+    ///
+    /// `sinkTask` is unsuitable here: it reuses one task identifier per
```
martinmitrevski, line 605: do we need this much details for a private method?

martinmitrevski (approved):
✅

Sources/StreamVideo/CallState.swift
ipavlidakis: I added it mainly to restrain myself from attempting to change it the next time I will see it 😅 I'm planning on fixing the sinkTask anyways in a follow up pr and all this comment will be removed anyways. If you can't stand it though I will happily remove it :)

PR #1162, opened 2026-06-12
[Fix] Fulfil CallKit answer action only after the call has successful…

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1759/blinkfix-answer-call-action-fulfilment-in-callkit.

### 🎯 Goal

The CXAnswerCallAction was fulfilled right after `accept()`, before the call had actually joined. This made the system UI replace "connecting…" with a running call duration prematurely, and handed audio session ownership to the SDK earlier than intended. The answer action is now fulfilled only once the call has successfully joined (join interceptor included).

### 📝 Summary

- Removed the premature `action.fulfill()` after `accept()`; the action is now fulfilled by the Call joining stage via the `JoinSource.ActionCompletion` hook, after the join (and any `CallJoinIntercepting`) has completed and right before the call becomes active.
- Exactly-once completion guard: on join failure the action is failed; if the failure surfaces after the action was already fulfilled (delivery-timeout race), the call is reported to CallKit as `.failed` instead.
- Implemented `provider(_:timedOutPerforming:)`: if CallKit times out the pending answer (slow join / long interceptor), the SDK aborts the join and leaves with reason `callkit.join.timeout`.
- Skipped the audio-session readiness wait for CallKit joins in the WebRTC joining stage (it became circular with the new ordering); peer connections are now configured before session activation — the canonical CallKit+WebRTC pattern.
- Fixed the resulting dead-mic regression with a new `CallKitRecoveryMiddleware`: when `.callKit(.activate)` arrives while recording is already on, the recording path is restarted (same recovery pattern as audio interruptions), so the mic is live without a manual mute toggle.
- Updated DocC across the lifecycle (`CallKitService`, `JoinSource.ActionCompletion`, joining stage, `configureAudioSession`) and added an "Answering calls" section to the CallKit integration docs page.

### 🛠 Implementation

- The joining stage invokes the `.callKit` completion inside the existing MainActor block that publishes `activeCall` — success path only; failure paths complete the action exactly once through an atomic guard in `CallKitService`.
- Audio hand-off ordering: `configureAudioSession(shouldSetActive: false)` → join completes → action fulfilled → CallKit activates the session → `didActivate` dispatches `.callKit(.activate)` into RTCAudioStore → deferred activation applies the audio policy. The recovery middleware reacts to the same action and restarts recording only when it was already on (in-app flows never dispatch it).
- Playout intentionally excluded from the restart: it is already refreshed via `audioSessionDidActivate` and the deferred `setAudioEnabled(true)`; the regression was capture-only.

### 🧪 Manual Testing Notes

All scenarios on a physical device, callee side answering via CallKit unless stated.

**Audio (peer connections now configured before session activation)**
1. Lock the device, receive a ringing call, answer from the lock screen (app cold-launched in background). Verify: remote audio audible promptly, and the local mic is live **without toggling mute** (caller hears you immediately) — regression check for the dead-mic fix.
2. Repeat with a Bluetooth headset connected — verify the route is honored once audio starts.
3. Repeat with the app in the foreground (in-app CallKit banner) — same audio expectations.

**System UI / answer-action lifecycle**
4. Answer with a normal network: system UI shows "connecting…" until the call joins, then the call duration starts (no premature duration).
5. Answer with a poor network (Network Link Conditioner): if the join completes within CallKit's answer window, the call connects normally; no premature duration at any point.
6. Enable the Synchronised Call Join Interceptor in the DemoApp debug menu and make the readiness handshake exceed CallKit's answer window (~5 s): the system ends the call and the SDK leaves with reason `callkit.join.timeout` (verify in logs); the app returns to a clean idle state and can receive the next call.
7. Answer, then hang up immediately while still "connecting…" — verify clean teardown.

**Audio session contention**
8. While on a regular phone (PSTN) call, answer a Stream call: if audio activation is delayed beyond ~10 s the audio watchdog forces a rejoin — verify the call recovers audio or fails cleanly, with no endless rejoin loop.

**CallKit mute sync**
9. After connecting, toggle mute/unmute from the CallKit system UI and from the in-app UI — verify both stay in sync and audio capture follows.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CallKit system UI no longer displaying running call duration before the user connects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Added support for SFU-provided WebRTC degradation preferences for outgoing video and screen-share tracks.[#1153](https://github.com/GetStream/stream-video-swift/pull/1153)
 - Exposed `callJoinInterceptor` on `CallKitAdapter` and `CallKitService` so CallKit-answered calls honor `CallJoinIntercepting`. [#1160](https://github.com/GetStream/stream-video-swift/pull/1160)
 
+### 🐞 Fixed
+- The CallKit system UI no longer shows a running call duration before the user is actually connected. [#1162](https://github.com/GetStream/stream-video-swift/pull/1162)
+
 # [1.47.1](https://github.com/GetStream/stream-video-swift/releases/tag/1.47.1)
 _May 26, 2026_
 
--- a/DocumentationTests/DocumentationTests/DocumentationTests/05-ui-cookbook/10-call-join-interceptor.swift
+++ b/DocumentationTests/DocumentationTests/DocumentationTests/05-ui-cookbook/10-call-join-interceptor.swift
@@ -97,9 +97,10 @@ private func content() {
             func callReadyToJoin(_ call: Call) async throws { /* Readiness work goes here. */ }
         }
 
-        // Assign the interceptor on the adapter so calls answered through
-        // CallKit honor it too.
+        // Assign the interceptor at startup, where streamVideo is handed to
+        // the adapter, so CallKit-answered calls honor it even on cold launch.
         @Injected(\.callKitAdapter) var callKitAdapter
+        callKitAdapter.streamVideo = streamVideo
         callKitAdapter.callJoinInterceptor = ParticipantReadyCallJoinInterceptor(streamVideo: streamVideo)
     }
 }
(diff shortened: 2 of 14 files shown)


martinmitrevski (approved):
lgtm ✅ let's test this a bit this week

Issue #1164, opened 2026-06-17
Audio does not recover after AVAudioSession interruption in Audio Rooms


afnanalfazary:
## What did you do?

1-  In an audio_room call with audio output enabled (listener mode), I let another app interrupt the audio session:
 Join an audio_room call and enable audio output using call.speaker.enableAudioOutput().
2- Background the app and start playing audio in another app (e.g. a video, podcast, or music app). iOS deactivates our AVAudioSession, and the room audio correctly becomes silent.
3. Stop the external audio and try to resume room audio either by: Tapping Play on the system Now Playing widget (which calls call.speaker.enableAudioOutput()), or  Returning the app to the foreground.

**The same behavior was reproduced in the official Stream demo application.**

## What did you expect to happen?

The call audio should resume normally. enableAudioOutput() should reactivate the audio session, restart playout, and the call should regain ownership of the Now Playing widget.

## What happened instead?
Audio never resumes.
call.speaker.enableAudioOutput() appears to be a no-op, the Now Playing widget remains owned by the other app, and the only way to recover audio is to leave the call and rejoin, which rebuilds the audio pipeline from scratch.

## GetStream Environment
**GetStream Video version:**  1.47.1
**GetStream Video frameworks:** StreamVideo, StreamVideoSwiftUI
**iOS version:** Reproduced on iOS 26.5 and iOS 18.6
**Swift version:** Swift 6
**Xcode version:**  Xcode 26
**Device:** iPhone 11, iPhone 13, iPhone 14 Pro Max

ipavlidakis:
Hi @afnanalfazary,

Thanks for reaching out. The `enableAudioOutput` is a `weird` API. We have ways to achieve similar things without using it. Firstly i need from you:
- Your code, so i can see what's going
- SDK logs (by enabled LogConfig.level = .debug). Make sure to provide the logs non-trimmed (it's fine to remove your API key but do not remove log lines) in a txt format (not pdf, rich-text or anything else)

Best regards
Ilias

afnanalfazary:
Hi @ipavlidakis 
Could you please share your email address so I can send the requested files, logs, and sample code?

ipavlidakis:
ilias.pavlidakis@getstream.io

afnanalfazary:
I tested the PR branch multiple times, including keeping the app interrupted/in the background for around 1 hour. In those scenarios the issue appears to be fixed and everything worked well — thanks so much for addressing it.

[...]

PR #1174, opened 2026-06-24
[Fix] Recover microphone after an interruption ends while backgrounded

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1773/gh-issueaudio-does-not-recover-after-avaudiosession-interruption-in
GH issue: https://github.com/GetStream/stream-video-swift/issues/1164

### 🎯 Goal

In Audio Rooms (and any call), the microphone could go permanently silent after the app spent time in the background with an audio interruption (e.g. another app playing audio). The participant appears unmuted and the SDK reports `isRecording: true`, but no audio is captured — and only leaving and rejoining the call restores it.

### 📝 Summary

- When recovering from an `AVAudioSession` interruption while muted-speech-detection is active, `InterruptionsEffect` now momentarily drops persistent recording mode around the recording restart, forcing WebRTC to actually restart the engine input instead of treating it as a no-op runtime toggle.
- Keeps the behaviour inside the existing `InterruptionsEffect` (no new type, no app-lifecycle coupling).
- Adds unit coverage asserting the dispatched recovery sequence for the muted-speech-detection path, while leaving the non-muted path unchanged.

### 🛠 Implementation

Root cause (confirmed against a customer log + the WebRTC fork):

- While the user is muted, the SDK enables muted-speech-detection, which puts WebRTC's ADM into persistent recording mode (`setRecordingAlwaysPreparedMode(true)` → `input_enabled_persistent_mode = true`).
- On interruption end, `InterruptionsEffect` restarts recording (`setRecording(false)` → `setRecording(true)`). But because persistent mode keeps `IsInputEnabled()` `true` across the stop/start, `DidUpdateAudioGraph()` is `false`, so `IsEngineRestartRequired()` is `false`. WebRTC applies it as a *runtime voice-processing mute update* and never restarts the input audio unit that the OS tore down during the interruption.
- Result: the engine reports `IsInputRunning: 1 / InputMuted: 0`, but the record device buffer delivers zero callbacks (`[REC` log stops at the interruption and never returns, while `[PLAY` keeps flowing). Playout works, capture is dead.
- `RTCAudioSession` re-delivers `didEndInterruption` on `handleApplicationDidBecomeActive`, so the recovery already runs in the foreground — but it still no-ops because of the masking above.

Fix:

- In `RTCAudioStore+InterruptionsEffect.didEndInterruption`, when `isRecording` and `isMutedSpeechDetectionEnabled`, dispatch `setMutedSpeechDetectionEnabled(false)` before `setRecording(false)/(true)` and `setMutedSpeechDetectionEnabled(true)` after. Disabling persistent mode makes the stop genuinely disable the input graph, so the subsequent start flips `IsInputEnabled` `false → true`, triggering a real engine restart and resurrecting capture.
- When muted-speech-detection is off (e.g. interrupted while unmuted), the sequence is unchanged — the normal stop/start already forces a restart there.

### 🎨 Showcase

_N/A — no UI changes._

| Before | After |
| ------ | ----- |
| Mic stays silent after a backgrounded interruption until leave/rejoin | Mic recovers automatically on returning to the foreground |

### 🧪 Manual Testing Notes

1. Join an Audio Room as a speaker (or any call while publishing audio) and mute yourself.
2. Background the app and play audio in another app (music/podcast/video) to trigger an `AVAudioSession` interruption; stop it / leave the app backgrounded for a bit.
3. Return to the foreground and unmute, then confirm:
   - Other participants can hear you again without leaving/rejoining.
   - WebRTC logs a `[REC : …]` device-buffer line again after the interruption (capture resumed) and outbound audio `packetsSent` increments.
   - `activeSessionIdentifier` does not change (no implicit leave/rejoin).
4. Regression: repeat background ⇄ foreground while unmuted and confirm the mic keeps working.

Automated: `RTCAudioStore_InterruptionsEffectTests` covers the recovery action sequence, including `test_didEndInterruption_shouldResumeTrue_recordingAndMutedSpeechDetection_togglesPersistentModeAroundRestart` and the unchanged non-muted path.

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### 🐞 Fixed
 - Local video could fail to start when joining a call if the camera capture session stopped unexpectedly (e.g. after a capture-server connection loss), because the capturer did not restart it. [#1175](https://github.com/GetStream/stream-video-swift/pull/1175)
+- The microphone now recovers after an audio-session interruption ends while the app is backgrounded, instead of staying silent (recording but capturing nothing) until you leave and rejoin the call. [#1174](https://github.com/GetStream/stream-video-swift/pull/1174)
 - Local video sometimes failed to start when joining a call, because the publisher negotiation was debounced on the main run loop and could be delayed while the call UI was presenting.[#1173](https://github.com/GetStream/stream-video-swift/pull/1173)
 - Constant audio route changes while joining a call no longer mute the microphone, so captured audio reaches the published track. [#1172](https://github.com/GetStream/stream-video-swift/pull/1172)
 
--- a/Sources/StreamVideo/Utils/AudioSession/RTCAudioStore/Namespace/Effects/RTCAudioStore+InterruptionsEffect.swift
+++ b/Sources/StreamVideo/Utils/AudioSession/RTCAudioStore/Namespace/Effects/RTCAudioStore+InterruptionsEffect.swift
@@ -57,10 +57,25 @@ extension RTCAudioStore {
                     state.audioDeviceModule != nil {
                     let isRecording = state.isRecording
                     let isMicrophoneMuted = state.isMicrophoneMuted
+                    let isMutedSpeechDetectionEnabled = state.isMutedSpeechDetectionEnabled
 
                     if isRecording {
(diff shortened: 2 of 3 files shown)


martinmitrevski (approved):
lgtm

PR #1178, opened 2026-07-02
[Enhancement]Migrate sdk signaling core infra onto stream core swift

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1816/migrate-sdk-signaling-and-core-infra-onto-stream-core-swift-websocket

### 🎯 Goal

StreamVideo duplicates a lot of low-level infrastructure that already lives in `stream-core-swift` (WebSocket client, logger, `APIError`, connection recovery). This PR routes both signaling WebSockets (SFU + coordinator), the logger, the API error model, and connection recovery through StreamCore instead of our own copies — removing duplication, keeping behavior consistent across Stream SDKs, and letting us track upstream fixes. It's the first, foundational step of the broader "converge StreamVideo on StreamCore" effort.

### 📝 Summary

- Add `stream-core-swift` (from `0.8.0`) as a dependency of `StreamVideo`, `StreamVideoSwiftUI`, and `StreamVideoUIKit`.
- **SFU signaling** now runs on `StreamCore.WebSocketClient` via a new `SFUWebSocket` isolation wrapper; state exposed as a video-side `SFUConnectionState`.
- **Coordinator signaling** now runs on `StreamCore.WebSocketClient` via a new `CoordinatorWebSocket` wrapper (auth handshake, event bridging, `connectionId`, state mapping).
- **Logger** unified on StreamCore (video's duplicated logger deleted, re-exposed via typealiases).
- **API errors** use `StreamAPIError` (`StreamCore.APIError`) across hand-written code; generated OpenAPI layer keeps its own `APIError`.
- **Reconnection** uses `StreamCore.DefaultConnectionRecoveryHandler`, with the video-specific CallKit policy preserved.
- Deleted video's now-dead WebSocket stack (client, engine, ping controller, recovery handler, providers) + their tests/mocks.

### 🛠 Implementation

**Isolation-wrapper strategy.** Importing StreamCore module-wide would collide with video's still-duplicated leaf types (`ClientError`, `DisposableBag`, `Event`, `EventNotificationCenter`, and — before this PR — the logger). To keep the blast radius small, each socket is wrapped in a single StreamCore-importing type (`SFUWebSocket`, `CoordinatorWebSocket`) that exposes a StreamCore-free, video-typed surface. Consumers (`SFUAdapter`, the WebRTC state machine, `StreamVideo`) stay StreamCore-free.

- **SFU:** `SFUWebSocket` owns `StreamCore.WebSocketClient`, maps `StreamCore.WebSocketConnectionState → SFUConnectionState`, bridges SFU payloads (`SFUEvent+StreamCore`, `WebRTCEventDecoder → StreamCore.AnyEventDecoder`). `SFUAdapter` is no longer a `ConnectionStateDelegate`; it mirrors state via the wrapper's publisher. A `SFUWebSocketProtocol` + `MockSFUWebSocket` keep it testable.
- **Coordinator:** `CoordinatorWebSocket` performs the `WSAuthMessageRequest` auth handshake on connect, decodes via video's `JsonEventDecoder` boxed as a `StreamCore.Event` (`CoordinatorEvent`) and forwards into video's existing `EventNotificationCenter`/middlewares (so event handling is unchanged), maps state back to video's `WebSocketConnectionState`, and exposes `connectionId`. `StreamVideo`/`StreamVideoEnvironment` were rewired onto it; token-refresh-on-invalid-token stays in `StreamVideo`.
- **Logger:** deleted `Publisher+Logger`/`Logger`/`LogSubsystem`/`LogLevel`/`LogConfig`/destinations/formatters/`Array+Logger`; re-exposed StreamCore's via module-wide typealiases (`StreamCoreLogging.swift`). Re-added `LogSubsystem.webRTCInternal` + `LogConfig.webRTCLogsEnabled`. `LogSubsystem.videoDefault` works around Swift's default-argument-value rule (member must be same-module).
- **Errors:** `StreamAPIError = StreamCore.APIError`. The transport decodes/throws it; `ClientError.apiError` is `StreamAPIError?`; token/client-error gating reads it. Only stored properties are accessed through the alias, so no StreamCore import is needed at call sites.
- **Reconnection:** `CoordinatorWebSocket` builds `StreamCore.DefaultConnectionRecoveryHandler`. StreamCore's own policies are `internal`, so the policy set is re-implemented against StreamCore's public APIs (extracted to `Coordinator/Reconnection/`) to preserve `internet AND wsAuto AND (background OR CallKit)`. CallKit's active-call check is injected as a closure (DI is ambiguous inside a StreamCore-importing file).

**Follow-ups are tracked in code** via a greppable `[StreamCore migration]` marker (reconnection policies once StreamCore makes them `public`; wrapper/alias/boxing removal once the event system + leaf types are unified; `StreamAPIError` removal once codegen emits `StreamCore.APIError`; restoring WebRTC log-severity sync). No public API is removed source-incompatibly; logger/error/state types are preserved by name via typealiases.

### 🧪 Manual Testing Notes

- Connect/disconnect a coordinator session; verify events (call updates, ring events) still arrive and `connectionId` is populated.
- Force a reconnection (toggle network / background during a call); verify the coordinator socket recovers, including **backgrounded CallKit calls** (CallKit reconnection policy).
- Trigger an expired/invalid token; verify token refresh + reconnect.
- Join a call and exercise SFU signaling (publish/subscribe, mute toggles, ICE restart) to confirm the SFU socket behaves as before.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved WebSocket connection handling, reconnection behavior, and CallKit-aware recovery.
  * Added more consistent server error reporting, including token expiration and client-error detection.
  * Unified SDK logging and WebRTC diagnostics under the shared logging system.
  * Improved SFU event and health-check processing.

* **Breaking Changes**
  * Legacy logging and WebSocket APIs have been replaced by shared SDK interfaces.
  * Some custom logging destination, formatter, and publisher helpers are no longer available.


Sources/StreamVideo/Utils/Logger/Data+DebugPrettyPrintedJSON.swift
```diff
@@ -0,0 +1,28 @@
...
+
+// Relocated here from the deleted `Logger.swift` during the StreamCore logger
+// migration. Kept as a video-internal extension (StreamCore also declares a
+// public `Data.debugPrettyPrintedJSON`): keeping ours out of any
+// StreamCore-importing file avoids an ambiguous member at the call sites that
+// use it without importing StreamCore.
+extension Data {
+    /// Converts the data into a pretty-printed JSON string. Use only for debug
+    /// purposes since this operation can be expensive.
+    var debugPrettyPrintedJSON: String {
```
martinmitrevski: why maintain two of them? Are they that much different?

Sources/StreamVideo/Utils/Extensions/Concurrency/Task+DisposableBag.swift
```diff
@@ -28,7 +28,10 @@ extension Task {
         disposableBag: DisposableBag,
         identifier: String = UUIDProviderKey.currentValue.get().uuidString,
         priority: TaskPriority? = nil,
-        subsystem: LogSubsystem = .other,
+        // `.videoDefault` (== `.other`) is a StreamVideo-owned member, so it can
```
martinmitrevski: I don't think these comments bring value tbh

Sources/StreamVideo/Errors/StreamAPIError.swift
```diff
@@ -0,0 +1,23 @@
...
+/// `import StreamCore` (which reintroduces `log`/`ClientError`/`APIError`
+/// ambiguities) in every consumer. Only stored properties (`code`,
+/// `statusCode`, `message`, `unrecoverable`) are read through it, so no import
+/// is needed at the use sites.
+///
+/// - TODO: [StreamCore migration] This alias (and the split between hand-written
+///   `StreamAPIError` and the generated `APIError`) goes away once the OpenAPI
+///   generator is changed to emit `APIError` as `StreamCore.APIError`. Then the
+///   whole app uses one API error type and this file can be deleted.
+public typealias StreamAPIError = StreamCore.APIError
```
martinmitrevski: can't we use the APIError directly?

Sources/StreamVideo/Utils/Logger/StreamCoreLogging.swift
```diff
@@ -0,0 +1,35 @@
...
+//
+
+import Foundation
+import StreamCore
+
+// StreamVideo's logging types are provided by StreamCore. These aliases expose
+// them under their historical StreamVideo names so existing call sites keep
+// working module-wide without importing StreamCore in every file.
+
+public typealias LogLevel = StreamCore.LogLevel
```
martinmitrevski: can't we solve this with @_exported import StreamCore?

martinmitrevski (commented):
partial review, will proceed at some point again

Sources/StreamVideo/WebRTC/v2/StateMachine/Stages/WebRTCCoordinator+Disconnected.swift
```diff
@@ -70,8 +70,10 @@ extension WebRTCCoordinator.StateMachine.Stage {
                 )
             } else if let error = context.flowError {
                 // If we got here from an unrecoverable error, we won't attempt
-                // any reconnection and instead we will disconnect.
-                if let apiError = error as? APIError, apiError.unrecoverable == true {
+                // any reconnection and instead we will disconnect. Errors are now
+                // `StreamAPIError` (StreamCore's); `unrecoverable` is a stored
+                // property, so no StreamCore import is needed here.
```
martinmitrevski: some of these comments are really not necessary.

Sources/StreamVideo/WebSockets/Client/Coordinator/Reconnection/BackgroundStateReconnectionPolicy.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+
+import Foundation
+import StreamCore
+
+/// Allows reconnection only while the app is active (foreground).
+///
+/// - TODO: [StreamCore migration] Delete this and reuse StreamCore's own
+///   `BackgroundStateReconnectionPolicy` once StreamCore makes it `public`.
+///   Currently `internal`, so re-implemented here. Logic is identical.
```
martinmitrevski: we can already make this public in core

Sources/StreamVideo/WebSockets/Client/Coordinator/Reconnection/CompositeReconnectionPolicy.swift
```diff
@@ -0,0 +1,32 @@
...
+//
+
+import Foundation
+import StreamCore
+
+/// Combines child policies with a logical `and`/`or`.
+///
+/// - TODO: [StreamCore migration] Delete this and reuse StreamCore's own
+///   `CompositeReconnectionPolicy` once StreamCore makes it `public`.
+///   Currently `internal`, so re-implemented here. Logic is identical.
```
martinmitrevski: same here, we can already make it public. All the changes that require changing to public are fine straight away, the others - let's review and discuss them.

Sources/StreamVideo/WebRTC/v2/SFU/SFUConnectionState.swift
```diff
@@ -0,0 +1,72 @@
...
+/// The connection state of the SFU signaling WebSocket, as exposed to the
+/// WebRTC state machine.
+///
+/// The SFU WebSocket is served by `StreamCore.WebSocketClient` (see
+/// `SFUWebSocket`). This video-side enum keeps StreamCore's connection-state
+/// type out of the WebRTC state machine (which uses many StreamVideo types and
+/// therefore cannot import StreamCore). `SFUWebSocket` maps
+/// `StreamCore.WebSocketConnectionState` into this type at the boundary.
+///
+/// - TODO: [StreamCore migration] This boundary enum exists so the WebRTC state
```
martinmitrevski: I don't really understand this reasoning, wdym by unifying leaf types?

Sources/StreamVideo/WebRTC/v2/SFU/SFUWebSocket.swift
```diff
@@ -0,0 +1,150 @@
...
+
+/// Owns the SFU signaling WebSocket, backed by `StreamCore.WebSocketClient`.
+///
+/// This is the single boundary that imports StreamCore for the SFU WebSocket.
+/// It exposes a StreamCore-free, StreamVideo-typed surface (`SFUConnectionState`,
+/// an SFU-payload event publisher, connect/disconnect/send) so `SFUAdapter` and
+/// the WebRTC state machine never import StreamCore.
+///
+/// - TODO: [StreamCore migration] This isolation wrapper exists because video
+///   still owns duplicated leaf types (`log`, `ClientError`, `DisposableBag`,
```
martinmitrevski: are there any blockers for unifying them or is it just still not done?

martinmitrevski (commented):
looks good so far. There's still some potential to reduce the SDK size further.

Sources/StreamVideo/Utils/Logger/Data+DebugPrettyPrintedJSON.swift
ipavlidakis: Addressed. Removed the local implementation and now use StreamCore’s formatter.

Sources/StreamVideo/Errors/StreamAPIError.swift
ipavlidakis: There are currently two nominally different APIError types: the generated StreamVideo.APIError and StreamCore.APIError. Using bare APIError resolves to the generated model, not StreamCore’s type. StreamAPIError therefore disambiguates Core errors without requiring import StreamCore in every consumer, which would also introduce symbol collisions. ClientError.apiError remains the generated type for source compatibility, with Core errors converted at the boundary. The alias can be removed once the generated model migrates to StreamCore (which i'm not even sure if it's possible to happen as it will require backend to align the apiErrors between chat and video and potentially feeds).

Sources/StreamVideo/Utils/Logger/StreamCoreLogging.swift
ipavlidakis: An exported import would expose all of StreamCore downstream and introduce collisions with Video’s existing types. The focused aliases preserve compatibility without broadening the public module surface, so I’ve kept the current approach.

Sources/StreamVideo/WebSockets/Client/Coordinator/Reconnection/BackgroundStateReconnectionPolicy.swift
ipavlidakis: Addressed. The duplicate implementation was removed and StreamCore’s public policy is used.

Sources/StreamVideo/WebSockets/Client/Coordinator/Reconnection/CompositeReconnectionPolicy.swift
ipavlidakis: Addressed. The duplicate implementation was removed and StreamCore’s public policy is used.

Sources/StreamVideo/WebRTC/v2/SFU/SFUWebSocket.swift
ipavlidakis: There is no fundamental blocker; it is intentionally deferred to keep this migration focused. The wrapper still normalizes StreamCore connection states, errors, and events into the types consumed by SFUAdapter, and provides its test-mocking seam. Removing it requires migrating SFUAdapter, the WebRTC state machine, and their mocks to consume StreamCore types directly. I’ve updated the documentation to make this remaining work explicit.

Sources/StreamVideo/WebRTC/v2/SFU/SFUConnectionState.swift
ipavlidakis: Those are are the leaf types https://github.com/GetStream/stream-video-swift/pull/1195 (basically types that are primitive but they are being used as they are without extensions)

Sources/StreamVideo/Errors/StreamAPIError.swift
```diff
@@ -0,0 +1,23 @@
...
+/// This is the canonical API error type for hand-written code; the generated
+/// OpenAPI layer keeps its own `APIError`. A distinct alias name is required
+/// because the generated `APIError` already occupies the bare name in this
+/// module, and referencing `StreamCore.APIError` directly would force
+/// `import StreamCore` (which reintroduces `log`/`ClientError`/`APIError`
+/// ambiguities) in every consumer. Only stored properties (`code`,
+/// `statusCode`, `message`, `unrecoverable`) are read through it, so no import
+/// is needed at the use sites.
+///
+/// - TODO: [StreamCore migration] This alias (and the split between hand-written
```
martinmitrevski: can't we fix this todo already?

Sources/StreamVideo/Utils/RawJSON+StreamCore.swift
```diff
@@ -0,0 +1,75 @@
...
+
+import Foundation
+import StreamCore
+
+/// A JSON value with an unknown structure.
+public typealias RawJSON = StreamCore.RawJSON
+
+public extension RawJSON {
+    /// Extracts the wrapped value as the specified type, if possible.
+    func value<T>() -> T? {
```
martinmitrevski, line 10: shouldn't this go into StreamCore?

Sources/StreamVideo/StreamVideo.swift
```diff
@@ -769,32 +762,30 @@ public class StreamVideo: ObservableObject, @unchecked Sendable {
...
+                    do {
+                        guard let apiTransport = apiTransport as? URLSessionTransport else { return }
+                        self.tokenSubject.send(try await apiTransport.refreshToken())
+                        log.debug("user token updated, will reconnect ws")
+                        webSocketClient?.connect()
+                    } catch {
+                        log.error("Error refreshing token, will disconnect ws connection", error: error)
                     }
-                } else {
-                    connectionRecoveryHandler?.webSocketClient(client, didUpdateConnectionState: state)
```
martinmitrevski, line 797: we don't need this any longer?

Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
```diff
@@ -3,6 +3,7 @@
 //
 
 import Combine
+import enum StreamCore.RawJSON
```
martinmitrevski: do we need to be this specific? Why not the whole core?

martinmitrevski (commented):
Looks good, let's test it a bit before merging. Left few comments as well.

Sources/StreamVideo/Errors/StreamAPIError.swift
ipavlidakis: https://github.com/GetStream/stream-video-swift/pull/1208 we can :)

Sources/StreamVideo/Utils/RawJSON+StreamCore.swift
ipavlidakis: If chat or feeds not using it, there is no reason to move it down to Core, right?

Sources/StreamVideo/StreamVideo.swift
ipavlidakis: The connectionRecoveryHandler is now part of the CoordinatorWebsocket, which is now the delegate to the handler https://github.com/GetStream/stream-video-swift/pull/1178/changes#diff-66385429e716ff2693e908df6930e45f6baf756f7a9dc83caac3cc07065b1bd5R148

Sources/StreamVideoSwiftUI/CallView/VideoParticipantsView.swift
ipavlidakis: Because we aren't consuming everything from StreamCore directly yet (mainly now the problem is the APIError) we rely on typealiases. That means that in files that both StreamCore and StreamVideo types collide, they create ambiguation causing compilation errors. The specific import is to avoid this problem. Once the APIError migration is in we can go through those specific imports and check if we can remove them but for now that should be fine.

Sources/StreamVideo/StreamVideo.swift
ipavlidakis: Just to be sure i added also a test to ensure connection recovery in `CoordinatorWebSocket_Tests `

Sources/StreamVideo/WebSockets/Client/Coordinator/CoordinatorWebSocket.swift
```diff
@@ -0,0 +1,120 @@
...
+
+    // MARK: - ConnectionStateDelegate
+
+    func webSocketClient(
+        _ client: WebSocketClient,
+        didUpdateConnectionState state: WebSocketConnectionState
+    ) {
+        if case let .disconnected(source) = state,
+           let serverError = source.serverError,
+           serverError.isInvalidTokenError || serverError.isTokenExpiredError {
```
ipavlidakis, line 108: If the token is expired, it doesn't forward the event to the recoveryHandler so no reconnection is being triggered.

Sources/StreamVideo/WebSockets/Events/Event.swift
```diff
@@ -37,31 +29,23 @@ extension Event {
...
+                return HealthCheckInfo(
+                    connectionId: healthCheckEvent.connectionId
+                )
             }
             if case let .typeConnectedEvent(connectedEvent) = event {
                 return HealthCheckInfo(
-                    coordinatorHealthCheck: .init(
-                        cid: nil,
-                        connectionId: connectedEvent.connectionId,
-                        createdAt: connectedEvent.createdAt
```
ipavlidakis, line 57: The StreamCore.HealthcheckInfo has no createdAt field

Sources/StreamVideo/StreamVideo.swift
```diff
@@ -539,7 +522,7 @@ public class StreamVideo: ObservableObject, @unchecked Sendable {
             apiKey: apiKey.apiKeyString
         )
         if let connectURL = try? URL(string: Self.endpointConfig.wsEndpoint)?.appendingQueryItems(queryParams) {
-            webSocketClient = makeWebSocketClient(url: connectURL, apiKey: apiKey)
+            webSocketClient = makeWebSocketClient(url: connectURL)
```
ipavlidakis, line 525: We don't need to pass in the apiKey any more as it's embedded in the connectURL

Sources/StreamVideo/StreamVideo.swift
```diff
@@ -769,32 +747,30 @@ public class StreamVideo: ObservableObject, @unchecked Sendable {
...
-                        do {
-                            guard let apiTransport = apiTransport as? URLSessionTransport else { return }
-                            self.tokenSubject.send(try await apiTransport.refreshToken())
-                            log.debug("user token updated, will reconnect ws")
-                            webSocketClient?.connect()
-                        } catch {
-                            log.error("Error refreshing token, will disconnect ws connection", error: error)
-                        }
+            if let serverError = source.serverError,
+               serverError.isInvalidTokenError || serverError.isTokenExpiredError {
```
ipavlidakis, line 761: This is the supplementary part for the token refresh and the WebsocketReconnectionPolicy. During a websocket disconnection StreamVideo and the CoordinatorWebsocket are being informed. If the disconnection was due to tokenExpired, then the CoordinatorWebsocket does nothing while StreamVideo kicks in a token refresh.

Package.swift
```diff
@@ -23,26 +23,35 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-protobuf.git", exact: "1.30.0"),
-        .package(url: "https://github.com/GetStream/stream-video-swift-webrtc.git", exact: "145.8.0")
+        .package(url: "https://github.com/GetStream/stream-video-swift-webrtc.git", exact: "145.8.0"),
+        .package(url: "https://github.com/GetStream/stream-core-swift.git", branch: "develop")
```
martinmitrevski, line 27: we should fix a version here

martinmitrevski (approved):
looks good to me, let's test a bit more and release

testableapple (approved):

PR #1179, opened 2026-07-03
[Fix] Prevent audio crash from concurrent AudioDeviceModule access

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1819/crashavaudioengine-on-1480

### 🎯 Goal

Users started crashing after 1.48.0 with an uncaught `NSException` thrown while WebRTC enabled voice processing on the shared `AVAudioEngine`. The goal is to stop the crash without any WebRTC fork changes, since none were made when it started.

### 📝 Summary

- Serialize every engine-mutating operation on `AudioDeviceModule` (`setRecording`, `setPlayout`, `resetPlayout`, `setMuted`, `refreshStereoPlayoutState`, `setStereoPlayoutPreference`, `setMutedSpeechDetectionEnabled`, `reset`) through a single re-entrant lock.
- Add a regression test that fails if two engine operations run concurrently.

### 🛠 Implementation

`AudioDeviceModule` drove the underlying WebRTC `RTCAudioDeviceModule` (and therefore the shared `AVAudioEngine` graph) synchronously on whatever queue each caller happened to run on. Those callers live on independent queues — the store middleware (`setRecording`/`setMuted`), `StereoPlayoutEffect`'s own `OperationQueue` (`refreshStereoPlayoutState`), and stats collection. `AVAudioEngine` graph mutation is not thread-safe, so when two of these overlapped, `-[AVAudioIONode setVoiceProcessingEnabled:error:]` → `AVAudioEngineGraph::_Connect` threw an uncaught `NSException` and the app aborted.

1.48.0 made the latent race reliably reproducible: `CallKitRecoveryMiddleware` now forces a `resetPlayout()` plus a recording restart burst on every CallKit activation, which coincides with the route-change-driven `refreshStereoPlayoutState()` running on the stereo effect's queue.

The fix introduces a private re-entrant `RecursiveQueue` on `AudioDeviceModule` and wraps each engine-mutating method in `engineQueue.sync { ... }`, so no two engine operations can overlap regardless of the calling queue. The lock is recursive because `setMuted` calls `setRecording`. Delegate callbacks from WebRTC only touch Combine subjects and are intentionally left unlocked to avoid re-entering during a locked `source` call.

### 🧪 Manual Testing Notes

Join a call over CallKit repeatedly (accept from the system UI) while toggling the mic and changing audio routes (speaker/headphones). Previously this could abort with the `AVAudioEngine` voice-processing exception; it should now stay stable.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a rare audio crash caused by concurrent audio device updates from multiple threads.
  * Improved reliability by serializing audio engine graph mutations and related audio state changes to prevent overlapping updates.
  * Enhanced audio engine lifecycle management to safely retain the previous engine until it’s released.
* **Tests**
  * Added regression coverage ensuring concurrent engine mutations are properly serialized.
  * Added tests verifying engines are retained and released correctly when replaced.
* **Chores**
  * Updated the changelog to document the fix.


martinmitrevski (approved):
✅

PR #1184, opened 2026-07-07
[Fix] Prevent os_unfair_lock crash when replacing a ringing call

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1873/bugon-call-deinit

### 🎯 Goal

Starting a new outgoing ring call could crash the app with a recursive
`os_unfair_lock` abort (`EXC_BREAKPOINT`) inside Combine. This fixes that crash.

### 📝 Summary

- Keep the previous `StreamVideo.State.ringingCall` value alive until after the
  `@Published` setter releases its internal lock, so the old `Call` is no longer
  deallocated inside `PublishedSubject.send`.
- Add a regression test that reproduces the exact crash chain and verifies the
  previous call is still released (just safely, after the lock is released).

### 🛠 Implementation

`ringingCall` is `@Published`. When a new value is assigned while the previous
call's last strong reference lived in that same property, `PublishedSubject.send`
swapped the stored value **while holding its `os_unfair_lock`**. That released the
old `Call`, so `Call.deinit` ran synchronously inside the locked region. The
call's `OutgoingRingingController` owns a subscription to that same `$ringingCall`
publisher, so tearing it down called `PublishedSubject.disassociate(_:)`, which
tried to re-acquire the lock the current thread already held → recursive abort.

The fix adds a `didSet { _ = oldValue }` observer to `ringingCall`. Referencing
`oldValue` forces the compiler to retain the previous value in a local temporary
across the wrapped-value assignment, so its refcount stays ≥ 1 while the subject
holds its lock. `deinit` (and the subscription's `disassociate`) then run after
the setter returns and the lock is released. The explicit `_ = oldValue` is
required: an empty `didSet` lets the optimizer skip materializing `oldValue`,
which would silently reintroduce the bug.

### 🧪 Manual Testing Notes

Start an outgoing ring call, cancel/reject it (leaving it briefly retained only by
`state.ringingCall`), then start a new ring call. Before the fix this aborts with
`_os_unfair_lock_recursive_abort`; after the fix the previous call is released
cleanly and the new call rings.

The added test `StreamVideo_Tests.test_streamVideo_ringingCallReplacement_releasesPreviousCallRetainedOnlyByRingingCall`
crashes without the fix (reproducing the production stack frame-for-frame) and
passes with it.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash that could occur when a new ringing call replaced an existing ringing call.
  * Improved ringing-call handling so the previous call is released safely after the state update completes.
  * Added regression coverage to help prevent this issue from returning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ### 🐞 Fixed
+- Fixed a crash that could occur when starting a new ringing call replaced a previous one that was still ringing. [#1184](https://github.com/GetStream/stream-video-swift/pull/1184)
 - Fixed an audio crash that could occur when the audio device module was updated from multiple threads at the same time. [#1179](https://github.com/GetStream/stream-video-swift/pull/1179)
 - Local video could fail to start when joining a call if the camera capture session stopped unexpectedly (e.g. after a capture-server connection loss), because the capturer did not restart it. [#1175](https://github.com/GetStream/stream-video-swift/pull/1175)
 - The microphone now recovers after an audio-session interruption ends while the app is backgrounded, instead of staying silent (recording but capturing nothing) until you leave and rejoin the call. [#1174](https://github.com/GetStream/stream-video-swift/pull/1174)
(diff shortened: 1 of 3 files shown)


martinmitrevski (approved):
lgtm

PR #1206, opened 2026-07-21
[Fix] Prevent camera flip from reverting to front camera

### LINKED ISSUE #1199: Camera flip during an active call switches back to the front camera in 1.49.0
## What did you do?

1. Join an active 1:1 video call with the front camera enabled (`videoOn: true`).
2. Wait until the front-camera capture is running.
3. Call `call.camera.flip()` once to switch to the back camera.

```swift
try await call.camera.flip()
```

This is a regression: the exact same code works correctly on **1.48.0** and fails on **1.49.0**. No other code changed between the two runs — only the SDK version.

The resolved `stream-video-swift` revisions were:

- 1.48.0: `eab7c4eb73b9f242af6a1991cbb21413ba3cc6a0`
- 1.49.0: `7890f5a8f9fede2f1932c1a50276c2544aee6fc2`

`stream-video-swift-webrtc` remained on 145.8.0 in both runs.

## What did you expect to happen?

The capture session switches to the back camera and stays there.

## What happened instead?

The back camera comes up for ~0.5s and then the capture session switches **back to the front camera** on its own. `call.camera.direction` stays `.back`, so SDK state and the actual capture device end up out of sync.

The logs and the 1.49.0 source point to the new `CameraInterruptionsHandler` added in [#1175](https://github.com/GetStream/stream-video-swift/pull/1175), "Local video could fail to start when joining a call if the camera capture session stopped unexpectedly."

To swap devices, [`CameraCaptureHandler.execute(...)`](https://github.com/GetStream/stream-video-swift/blob/1.49.0/Sources/StreamVideo/WebRTC/v2/VideoCapturing/ActionHandlers/Camera/CameraCaptureHandler.swift#L140-L145) deliberately stops the running capture session:

```swift
// If the capturer is already active, since iOS 18, we need to explicitly
// stop the current capturing session before we start one with the new
// configuration.
if activeConfiguration != nil {
    await videoCapturer.stopCapture()
}
```

That stop posts `AVCaptureSession.didStopRunningNotification`, which [`CameraInterruptionsHandler`](https://github.com/GetStream/stream-video-swift/blob/1.49.0/Sources/StreamVideo/WebRTC/v2/VideoCapturing/ActionHandlers/Camera/CameraInterruptionsHandler.swift#L119-L125) observes and misreads as an unexpected stop:

```swift
NotificationCenter
    .default
    .publisher(for: AVCaptureSession.didStopRunningNotification, object: session)
    .receive(on: processingQueue)
    .sink { [weak self] _ in self?.attemptRestart(reason: "session stopped unexpectedly") }
```

`attemptRestart` then re-dispatches the cached `lastStartAction` — which still holds the **join-time** `.startCapture(position: .front)`, because `handle(_:)` only updates `lastStartAction` on `.startCapture` and ignores `.setCameraPosition` entirely (`default: break`). The recovery therefore restarts the front camera and undoes the flip.

The handler's own doc comment says intentional stops should not be misread:

> The interruptions handler is placed before the capture handler so it updates its state and observers before the capture session is started or stopped. This keeps an intentional `stopCapture` from being misread as an unexpected stop (...)

This holds for the `.stopCapture` **action** (which calls `didStopCapture()` → `state = .idle`), but `.setCameraPosition` stops the session *inside* the capture handler without ever emitting a `.stopCapture` action, so the interruptions handler stays `.running` and fires.

Device log of a single flip tap (front → back), trimmed to the relevant lines:

```
20:18:14.998 [WebRTCCoordinator.swift:173] [changeCameraMode(position:)] > CallSettings updated <... cameraPosition:front/> -> <... cameraPosition:back/>
20:18:15.008 (RTCCameraVideoCapturer.m:264 stopCaptureWithCompletionHandler): Stop
20:18:15.078 (RTCCameraVideoCapturer.m:460 handleCaptureSessionDidStopRunning): Capture session stopped.
20:18:15.079 [WARNING] [CameraInterruptionsHandler.swift:223] [attemptRestart(reason:)] > CameraCapture needs recovery (session stopped unexpectedly). Restarting capture (attempt 1).
20:18:15.347 [CameraCaptureHandler.swift:157] > CameraCaptureHandler started capturing with configuration position:.back dimensions:(1280.0, 720.0) frameRate:30.
20:18:15.354 (RTCCameraVideoCapturer.m:264 stopCaptureWithCompletionHandler): Stop
20:18:15.481 [CameraCaptureHandler.swift:263] [executeStop(_:)] > CameraCaptureHandler stopped capturing.
20:18:15.635 [CameraCaptureHandler.swift:157] > CameraCaptureHandler started capturing with configuration position:.front dimensions:(1280.0, 720.0) frameRate:30.
```

`14.998` is the flip, `15.347` is the successful switch to `.back`, and `15.635` is the spurious recovery restarting `.front`.

Confirmed by diffing the tags: `CameraInterruptionsHandler.swift` at 1.48.0 only observes `wasInterrupted` / `interruptionEnded` and has no `didStopRunning` observer, no `attemptRestart`, and no `lastStartAction` — which is why flip works there.

## GetStream Environment

- **GetStream Video version:** 1.49.0 (regression; 1.48.0 works)
- **GetStream Video frameworks:** StreamVideo, StreamVideoSwiftUI
- **iOS version:** 26.6.2 (iPhone 13), 16.7.16 (iPhone 8)
- **Swift version:** Apple Swift 6.3.1 (Swift language mode 5)
- **Xcode version:** 26.4.1 (17E202)
- **Device:** iPhone 13 and iPhone 8 — reproduced on both devices

### Possible fix

The position-change path appears to require both of the following:

1. Do not treat the intentional capture-session stop performed by `.setCameraPosition` as an unexpected stop requiring recovery.
2. After the new position is applied, update the cached recovery configuration so a genuine later recovery starts the current camera rather than the join-time camera.

### Attempted workaround

We also tried wrapping the flip in `disable()` → `flip()` → `enable()`.

This avoids the stale front-camera restart: the `.stopCapture` action disarms the interruptions handler, the subsequent `.startCapture` refreshes `lastStartAction`, and the back camera starts successfully. However, the camera is disabled again shortly afterward, so this is not a usable workaround in our environment:

```
50.941 [CameraCaptureHandler.swift:263] > CameraCaptureHandler stopped capturing.              <- disable()
51.350 [CameraCaptureHandler.swift:157] > started capturing with configuration position:.back  <- flip() + enable(), correct
51.537 [CameraCaptureHandler.swift:263] > CameraCaptureHandler stopped capturing.              <- disabled again
```

<details>
<summary>Logs and suspected ordering issue</summary>

`WebRTCStateAdapter.updateCallSettings(from: Stream_Video_Sfu_Event_TrackUnpublished)` sets `videoOn` to `false` for a `TrackUnpublished` event belonging to its own `sessionID`:

```swift
guard event.participant.sessionID == sessionID, ...
enqueueCallSettings { currentCallSettings in
    switch event.type {
    case .video: return currentCallSettings.withUpdatedVideoState(false)
```

The timing and logs suggest that a delayed `TrackUnpublished` event for our own video track updates `videoOn` to `false` after the camera has already been re-enabled.

This appears to be a separate ordering issue and is included here only to explain why the obvious `disable()` → `flip()` → `enable()` mitigation is not viable. We can file it separately if needed.

</details>

We are pinning to 1.48.0 for now.

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1894/bugflip-camera-issue-on-1490
- Closes [#1199](https://github.com/GetStream/stream-video-swift/issues/1199)

### 🎯 Goal

Fix the camera-flip regression introduced in 1.49.0. An intentional capture-session stop while switching cameras was treated as an unexpected failure, causing recovery to restart the cached front-camera configuration after the back camera had already started.

The selected camera should remain active and stay aligned with `call.camera.direction`, while genuine unexpected capture-session stops should continue to recover normally.

### 📝 Summary

- Track intentional camera-position transitions across capture-session stop/start notifications.
- Keep the cached recovery action aligned with the latest requested camera position.
- Roll back staged transition state when a downstream capture-action handler fails.
- Cover single and consecutive flips, expected and unexpected stops, and rollback behavior with focused tests.

### 🛠 Implementation

`CameraInterruptionsHandler` now serializes capture actions and session notifications on its existing processing queue. Camera-position changes enqueue an identity-bearing transition that moves from `awaitingStop` to `awaitingStart`; matching notifications consume transitions in FIFO order. Automatic recovery remains suppressed while intentional transitions are pending, including during rapid consecutive flips.

Handling `.setCameraPosition` also rebuilds the cached start action with the requested position and current capture dependencies. A later genuine unexpected stop therefore restarts the camera using the latest successful position instead of the join-time position.

`StreamVideoCapturer` now invokes `handleFailure(for:)` on its action handlers when any handler throws. `CameraInterruptionsHandler` uses that callback to remove the exact staged transition and restore the previous recovery configuration before the original error is rethrown.

### 🧪 Manual Testing Notes

1. Run the DemoApp on a physical device.
2. Join an active video call with the front camera enabled.
3. Flip once to the back camera.
4. Verify the back camera remains active and `call.camera.direction` stays `.back`.
5. Flip repeatedly between front and back cameras and verify each selected camera remains active.

### ☑️ Contributor Checklist

- [ ] Comparison screenshots added for visual changes (not applicable — no visual changes)
- [ ] Affected documentation updated (DocC; no tutorial or CMS changes required)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery for capture-session interruptions and unexpected stops.
  * Prevented automatic restart during intentional camera position changes.
  * Ensured failed position changes roll back to the last successful camera position.
  * Fixed camera flips no longer reverting to the front camera during active calls.
* **New Features**
  * Added failure/rollback handling for action handlers so staged state can be restored when a handler fails.
* **Tests**
  * Added XCTest coverage for interruption, consecutive position changes, failed transitions, and restart sequencing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### 🔄 Changed
 
+### 🐞 Fixed
+- Fixed camera flips reverting to the front camera during active calls. [#1206](https://github.com/GetStream/stream-video-swift/pull/1206)
+
 # [1.49.0](https://github.com/GetStream/stream-video-swift/releases/tag/1.49.0)
 _July 08, 2026_
 
--- a/Sources/StreamVideo/WebRTC/v2/VideoCapturing/ActionHandlers/Camera/CameraInterruptionsHandler.swift
+++ b/Sources/StreamVideo/WebRTC/v2/VideoCapturing/ActionHandlers/Camera/CameraInterruptionsHandler.swift
@@ -28,20 +28,67 @@ final class CameraInterruptionsHandler: StreamVideoCapturerActionHandler, @unche
         case running(session: AVCaptureSession, disposableBag: DisposableBag)
     }
 
+    /// Tracks one intentional camera-position transition from its expected
+    /// capture-session stop until the corresponding start notification.
+    ///
+    /// Instances provide transition identity in addition to ordering. This lets
+    /// failure rollback remove the exact in-flight change without completing an
+    /// older or newer change during rapid consecutive flips.
+    private final class CameraPositionChange {
+        /// Notification the transition is waiting to receive next.
+        enum State: Equatable {
+            /// The capture handler has not stopped the previous device yet.
+            case awaitingStop
+            /// The expected stop arrived; the replacement device has not started yet.
+            case awaitingStart
+        }
+
+        /// Start action to restore if a later action handler rejects this change.
+        let previousStartAction: StreamVideoCapturer.Action
+        /// Current notification phase for this transition.
+        var state: State = .awaitingStop
+
+        init(previousStartAction: StreamVideoCapturer.Action) {
+            self.previousStartAction = previousStartAction
+        }
+    }
+
     /// Maximum number of consecutive automatic restart attempts before giving
     /// up, to avoid a restart loop when the capture server cannot recover. The
     /// counter resets once capture successfully starts again.
     private static let maxRestartAttempts = 3
 
-    /// Serializes notification handling. All observers deliver on this queue
-    /// (via `receive(on:)`), so the restart bookkeeping below is mutated from a
-    /// single place.
+    /// Serializes lifecycle actions, failure callbacks, and notifications.
+    ///
+    /// All observers deliver on this queue through `receive(on:)`, while
+    /// ``handle(_:)`` and ``handleFailure(for:)`` enqueue synchronous operations.
+    /// Keeping every transition mutation here prevents capture actions from
+    /// racing delayed `AVCaptureSession` notifications.
     private let processingQueue = OperationQueue(maxConcurrentOperationCount: 1)
     private var state: State = .idle
     private var lastStartAction: StreamVideoCapturer.Action?
     private var restartAttempts = 0
     private var isInterrupted = false
     private var isRestarting = false
+
+    /// Intentional camera changes awaiting their stop/start notifications.
+    ///
+    /// Notification handlers consume this collection in FIFO order because
+    /// `AVCaptureSession` posts lifecycle notifications in transition order.
+    private var cameraPositionChanges: [CameraPositionChange] = []
+
+    /// Change staged by the most recent position-change pipeline invocation.
+    ///
+    /// ``handleFailure(for:)`` matches this identity against the FIFO before
+    /// rolling back. Once notifications complete and remove the transition, a
+    /// retained reference cannot roll back an already completed change.
+    private var currentCameraPositionChange: CameraPositionChange?
+
+    /// Whether automatic recovery must remain suppressed for an intentional
+    /// camera stop/start cycle.
+    private var isChangingCameraPosition: Bool {
+        !cameraPositionChanges.isEmpty
+    }
 
     /// Dispatches actions back through the capturer pipeline. Assigned by
     /// ``StreamVideoCapturer`` so the handler can issue a full capture restart.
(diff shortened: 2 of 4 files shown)


martinmitrevski (approved):
looks good

PR #1207, opened 2026-07-21
[Fix] Outgoing ring not cancelling when the caller hangs up during call creation

ipavlidakis:
### 🔗 Issue Links

Partial Resolves IOS-1895

### 🎯 Goal

When a caller starts a ringing (outgoing) call and hangs up almost
immediately — before `create(ring:)` finishes — the cancellation could
reach the backend before the call existed. The `reject(cancel)` then
failed silently, so the callee kept ringing and could even answer into a
call the caller had already abandoned. This PR guarantees the cancel is
delivered once the call has actually been created.

### 📝 Summary

- Add `OutgoingRingCreationCoordinator`, a small serial arbiter that
  funnels the two competing outcomes of an outgoing ring — "create
  finished" and "user hung up" — so a hang-up that races creation is
  never lost.
- Wire `CallViewModel` start/hang-up through the coordinator: if the
  hang-up lands while `create(ring:)` is still in flight, the create path
  issues the cancel the moment the call exists.
- Add a deterministic regression test plus `MockCall` hooks that
  reproduce the race.

### 🛠 Implementation

- `OutgoingRingCreationCoordinator` guards a three-state latch
  (`creating` / `created` / `rejected`) behind a single `UnfairQueue`
  lock. `created()` and `rejected()` are synchronous and throw typed
  errors (`CallAlreadyRejected`, `CallNotCreatedYet`) so the cancel is
  handed to whichever side finishes last — turning an unlucky ordering
  into a deterministic hand-off.
- In `startCall(ring:)` the coordinator is created before `call.create`.
  On success, `created()` throws `CallAlreadyRejected` if the user
  already hung up, and the create path rejects with `.cancel`.
- In `handleCallHangUp`, `rejected()` throws `CallNotCreatedYet` when
  create is still running, so the hang-up defers the cancel to the create
  path instead of rejecting a call the backend does not know about yet.
- This coordinator governs only the creation-vs-cancel handshake;
  runtime concerns (e.g. auto-ending the ring on backgrounding) remain in
  `OutgoingRingingController`.

### 🧪 Manual Testing Notes

- Start an outgoing ringing call and hang up immediately(place your finger where the hang up button is and once you press the start call press multiple time the hang up), before it
  connects. Confirm the callee stops ringing and cannot answer into an
  empty call.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Hanging up an outgoing call immediately after starting it now reliably prevents the recipient from continuing to ring.
  - Improved handling of hang-up timing during outgoing call creation so cancellation/rejection completes cleanly without premature outcomes.

- **Tests**
  - Added a regression test for hang-up during outgoing call creation, verifying the rejection occurs after creation and uses the expected cancel reason.
  - Broadened automated coverage by removing the bundle-level skip for call view model tests.


Sources/StreamVideoSwiftUI/CallViewModel.swift
```diff
@@ -1003,7 +1033,13 @@ open class CallViewModel: ObservableObject {
 
         Task(disposableBag: disposableBag, priority: .userInitiated) { [weak self] in
             guard let self else { return }
+
             do {
+                // Tell the coordinator the user hung up. If create is still in
+                // flight this throws `CallNotCreatedYet`, and we defer the
+                // cancel to the create path instead of rejecting a call the
+                // backend does not know about yet.
+                try ringCreationCoordinator?.rejected()
```
martinmitrevski: shouldn't we set the coordinator to nil here as well? Otherwise, the call would still be alive.

Sources/StreamVideoSwiftUI/CallViewModel.swift
```diff
@@ -476,11 +489,28 @@ open class CallViewModel: ObservableObject {
                         maxParticipants: maxParticipants,
                         video: video
                     )
+
+                    // Create succeeded. Announce it to the coordinator: this
+                    // throws `CallAlreadyRejected` if the user hung up while we
+                    // were creating, so the ring timer below is only armed for
+                    // a call that is genuinely still ringing.
+                    try ringCreationCoordinator?.created()
```
martinmitrevski: I think it would've been easier to just sync the state directly in the call view model. This is a bit error prone as well, you need to make sure to nil it out always, which also feels weird - first you call `created`, then set it to nil.

Sources/StreamVideoSwiftUI/Utils/OutgoingRingCreationCoordinator.swift
```diff
@@ -0,0 +1,105 @@
...
+///
+/// The transitions are a synchronous compare-and-set, so a plain lock is
+/// enough — there is no work awaited while holding it. This governs only the
+/// creation-vs-cancel handshake; runtime concerns such as auto-ending the
+/// ring when the app backgrounds live in ``OutgoingRingingController``.
+final class OutgoingRingCreationCoordinator: @unchecked Sendable {
+    /// Thrown from ``created()`` when the call was hung up before it finished
+    /// being created. The create call site catches this and issues the cancel
+    /// itself, now that the call actually exists on the backend to be rejected.
+    struct CallAlreadyRejected: Error {}
```
martinmitrevski: shouldn't we also have them as `final class`?

martinmitrevski (commented):
I think we don't need the extra object, especially since we need to pass around the call.

Sources/StreamVideoSwiftUI/Utils/OutgoingRingCreationCoordinator.swift
ipavlidakis: You are probably worried about size i guess. Yes can make them `final class`

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: I didn't like adding another state control in the CallViewModel. It made following things harder and it was much easier to break it in the future as the CallViewModel has a far bigger interface to touch.

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: No we shouldn't as if the `rejected` fails it means that the call hasn't been created yet and the `ringCreationCoordinator` is required so that on line 497 (after the call creation) the coordinator will trigger the `created` which will again fail causing the creation flow to trigger the rejection on the backend.

ipavlidakis:
> I think we don't need the extra object, especially since we need to pass around the call.

The readability without the object was difficult. I also believe it will be more error prone if we just have it inside CallViewModel as there are more chances we may need to make changes on the ViewModel that can affect this code and we may miss any change there

Sources/StreamVideoSwiftUI/CallViewModel.swift
martinmitrevski: another reason why all this feels too complex

Sources/StreamVideoSwiftUI/CallViewModel.swift
martinmitrevski: I think if we just named it ringingTask or something, it's not going to be misused

martinmitrevski (commented):
I still think this object just adds unnecessary complexity without adding too much value. We move the call object to another object, which we sometimes nil out, sometimes we don't - which in turn depends on other factors. I think it's a bit hard to follow.

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: Just having the task stored and cancelling doesn't solve the problem. We need to know inside the task if the rejection has already occurred. This is why i got the coordinator as having a single source of truth for the rejection-during-call-creation. So it's not only a stored task it's also boolean that then becomes an enum that needs to be stored inside the CallViewModel (extending and make it more complex and error prone).

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: I'm not sure I agree with this one. The code has a single point of creating the coordinator and nullifying it. It has comments and very specific API. I don't see how that adds complexity in the fact that the coordinator may be nullified here. And in any case, ofc there is complexity, this is just patching the model internally. There are better ways but simply storing the task didn't resolve the problem and this why i got the coordinator that keeps track of the rejections and then inside the ringingTask we know if the call has been rejected so we can end it in a valid way.

ipavlidakis:
The coordinator has a very specific lifecycle. I'm not sure what is that is causing the confusion. As i mentioned in my comments above there is no comparison between having a stored task and the object as the store task itself didn't resolve the problem as we need to keep a state to know if the call has been rejected before creation and if yes inside the ringing task we shouldn't simply cancel but we should reject.

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: As we discussed this one is now updated so that the name reflects more the state management the object does internally.

Sources/StreamVideoSwiftUI/CallViewModel.swift
ipavlidakis: As we discussed this one is now updated so that the name reflects more the state management the object does internally.

ipavlidakis:
As we discussed this one is now updated and the Call object isn't necessary any more to instantiate the object .

martinmitrevski (approved):
lgtm

PR #1232, opened 2026-08-11
[Fix] Restore output route after media services reset

ipavlidakis:
### 🔗 Issue Links

[IOS-1961](https://linear.app/stream/issue/IOS-1961/bugmediaservices-teardown-issue)

### 🎯 Goal

Restore the SDK-selected audio output route after iOS restarts media services, without reactivating a session that CallKit or the app has already deactivated.

### 📝 Summary

- Publish WebRTC media-services reset callbacks through `RTCAudioSessionPublisher`.
- Dispatch route recovery through the existing audio-store pipeline.
- Reapply the cached output-port override for play-and-record sessions.
- Add unit coverage for the publisher, effect, coordinator, and reducer behavior.

### 🛠 Implementation

`RTCAudioSessionPublisher` converts `audioSessionMediaServerReset` delegate callbacks into a store event. `RouteChangeEffect` receives the event on its processing queue and dispatches a dedicated recovery action.

The coordinator always executes this action because the stored route configuration may be unchanged while the media-services restart has discarded the corresponding system state.

The reducer reapplies the cached `overrideOutputAudioPort` value only for `.playAndRecord` sessions. It does not activate the audio session, ensuring that a concurrent CallKit or application deactivation remains authoritative.

This restores the route state owned by the Swift SDK. Full bidirectional audio recovery also depends on the companion WebRTC change that restores `AVAudioSession` configuration and rebuilds the invalidated audio engine. This branch does not update the StreamWebRTC dependency.

### 🎨 Showcase

Not applicable — no UI changes.

### 🧪 Manual Testing Notes

Suggested physical-device scenario:

1. Integrate a StreamWebRTC build containing the companion media-services recovery.
2. Join a call and select speaker output.
3. Trigger a media-services reset while the call remains connected.
4. Verify the speaker override is restored and bidirectional audio resumes.
5. Repeat while leaving the call or deactivating through CallKit.
6. Verify the session remains inactive and audio is not restarted unexpectedly.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio-session recovery after media services reset.
  * Restores the configured speaker output for active play-and-record sessions without unnecessarily reactivating the audio session.
  * Helps maintain expected audio routing when iOS media services restart.

* **Compatibility**
  * Updated the underlying WebRTC integration to improve media-session reset handling.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -23,7 +23,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-protobuf.git", exact: "1.30.0"),
-        .package(url: "https://github.com/GetStream/stream-video-swift-webrtc.git", exact: "145.13.0"),
+        .package(url: "https://github.com/GetStream/stream-video-swift-webrtc.git", exact: "145.14.0"),
         .package(url: "https://github.com/GetStream/stream-core-swift.git", from: "0.10.1")
     ],
     targets: [
--- a/Sources/StreamVideo/Generated/SystemEnvironment+Version.swift
+++ b/Sources/StreamVideo/Generated/SystemEnvironment+Version.swift
@@ -9,5 +9,5 @@ extension SystemEnvironment {
   /// A Stream Video version.
   public static let version: String = "1.51.0-SNAPSHOT"
   /// The WebRTC version.
-  public static let webRTCVersion: String = "145.13.0"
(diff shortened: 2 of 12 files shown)


martinmitrevski (approved):
lgtm

PR #1234, opened 2026-08-12
Keep Float in generated OpenAPI models

gpunto:
### 🔗 Issue Links

- [CHA-4672](https://linear.app/stream/issue/CHA-4672/generated-sdk-models-narrow-float64-fields-to-32-bit-float)
- Depends on GetStream/chat#15645

### 🎯 Goal

The chat OpenAPI Swift generator now maps JSON numbers to `Double` instead of `Float`, since `Float` truncates the `float64` values the backend sends.

Our generated models are `public`, so widening them would be a source and binary break. Our number fields are also response-only, so nothing is sent back narrowed and the extra precision isn't consumed anywhere — `EdgeResponse` coordinates pick a datacenter, and `Float` is already sub-metre there.

### 📝 Summary

- Pass `number_as_float=true` in `Scripts/generate.sh` to keep the current types.
- No generated code changes: this offsets the new generator default exactly.
- The opt should be dropped at the next major to adopt `Double`.

### 🛠 Implementation

The width is a generator opt rather than a hardcoded exclusion, so adopting `Double` later is a one-line deletion here instead of a change in the chat repo. The generator rejects unknown opts, so a typo fails the generation run rather than silently widening.

### 🧪 Manual Testing Notes

Ran `Scripts/generate.sh` against the chat branch and diffed `Sources/StreamVideo/OpenApi/generated/` against a run on chat `master` without the opt: byte identical, 337 files.

Note that GetStream/chat#15645 must merge first — the opt does not exist before it.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Improvements**
  * Updated Swift client generation so JSON numbers are represented as `Float` values instead of `Double`.

### DIFF
--- a/Scripts/generate.sh
+++ b/Scripts/generate.sh
@@ -19,7 +19,10 @@ rm -rf "$PROJECT_ROOT/Sources/StreamVideo/OpenApi/generated/Models/"*
 (
   cd "$SOURCE_PATH/projects/chat-manager" &&
   go run . openapi generate-spec -products video -version v2 -clientside -output "$SOURCE_PATH/releases/v2/video-openapi-clientside" -renamed-models "$SCRIPT_DIR/renamed-models.json" &&
-  go run . openapi generate-client --language swift --spec "$SOURCE_PATH/releases/v2/video-openapi-clientside.yaml" --output "$PROJECT_ROOT/Sources/StreamVideo/OpenApi/generated/"
+  # number_as_float keeps JSON numbers as Float. The generator defaults to Double
+  # since Float truncates float64s, but our generated models are public, so
+  # widening them would be a source and binary break.
+  go run . openapi generate-client --language swift --spec "$SOURCE_PATH/releases/v2/video-openapi-clientside.yaml" --opt number_as_float=true --output "$PROJECT_ROOT/Sources/StreamVideo/OpenApi/generated/"
 )
 
 # Shared OpenAPI types are provided by StreamCore.


Scripts/generate.sh
```diff
@@ -19,7 +19,10 @@ rm -rf "$PROJECT_ROOT/Sources/StreamVideo/OpenApi/generated/Models/"*
 (
   cd "$SOURCE_PATH/projects/chat-manager" &&
   go run . openapi generate-spec -products video -version v2 -clientside -output "$SOURCE_PATH/releases/v2/video-openapi-clientside" -renamed-models "$SCRIPT_DIR/renamed-models.json" &&
-  go run . openapi generate-client --language swift --spec "$SOURCE_PATH/releases/v2/video-openapi-clientside.yaml" --output "$PROJECT_ROOT/Sources/StreamVideo/OpenApi/generated/"
+  # number_as_float keeps JSON numbers as Float. The generator defaults to Double
+  # since Float truncates float64s, but our generated models are public, so
+  # widening them would be a source and binary break.
+  go run . openapi generate-client --language swift --spec "$SOURCE_PATH/releases/v2/video-openapi-clientside.yaml" --opt number_as_float=true --output "$PROJECT_ROOT/Sources/StreamVideo/OpenApi/generated/"
```
ipavlidakis, line 25: What values is this really affecting? Do we have any field that the truncation (from double to float) can actually cause issues?

gpunto: So, I had Claude check the fields: 17 number fields in our spec (`-products video`) across 12 property names. lat/lng on `EdgeResponse` / `CoordinatesResponse` / `CallStatsLocation`, the scores and ratings (`score`, `avg_user_rating`, `average`), the four `NetworkMetricsReportResponse` ms averages, `distance_to_sfu_kilometers`, `MetricThreshold.value`, `Bound.value`, `ReportByHistogramBucket.sum`.

Nothing materially affected. Coordinate error measures 0.01–0.12 m (worst ~0.2 m) and we use them to pick a datacenter; the rest are 1–5 scores, ms latencies and distances, all well inside `Float`'s ~7 significant digits.

Also all 17 are response-only. `UpdateLiveLocationRequest` comes from `common` and isn't in our spec, so we never send a narrowed value back. The one theoretical exception is `ReportByHistogramBucket.sum`, where the integer part of the `Float` stops being exact above 16.7M.

Note these are the fields in the current spec, the generated code in the repo is stale (last regenerated in July, several of these `Float` fields don't exist in the spec anymore), so a regeneration will carry also an unrelated diff on top.

ipavlidakis: Ok, that sounds quite safe. Thanks for elaborating.

ipavlidakis (approved):

PR #1243, opened 2026-08-17
Expose Call.delete on iOS

martinmitrevski:
### 🔗 Issue Links

N/A — found while auditing the public API surface against the other Video SDKs.

### 🎯 Goal

The `deleteCall` coordinator endpoint has been present in our generated OpenAPI layer for a while, but nothing exposed it. Applications that need to delete a call — rather than just end the active session — currently have no way to do so on iOS.

### 📝 Summary

- Added `Call.delete(hard:)`, which wraps the already-generated `deleteCall` endpoint.
- Added unit tests covering the request payload, the returned response, and error propagation.
- Added a `DeleteCallResponse.dummy(...)` test helper, matching the existing dummy convention.
- Updated the changelog.

### 🛠 Implementation

`delete(hard:)` sits next to `end()` in `Call.swift`, since the two are related lifecycle operations.

Two decisions worth calling out:

- **`hard: Bool` parameter instead of a request object.** `DeleteCallRequest` has exactly one field, so nothing is lost by flattening it, and it reads more naturally at the call site. This matches how `goLive` and `update` take individual parameters rather than a request.
- **Thin pass-through — no state writes and no cache eviction.** This mirrors `end()`, which also leaves state to the corresponding event. The backend emits `CallDeletedEvent`, which `CallState` already handles, so writing the response back into state would only duplicate that path — and for a hard delete, restoring the deleted call's state locally would be actively misleading. The doc comment notes that participants still in the call are not detached automatically, so callers should `leave(reason:)` if the current user has joined.

Tests reuse the existing `MockDefaultAPIEndpoints` pattern from `test_kickUser_coordinatorWasCalledWithExpectedValues`. I extracted the shared `Call` + mock setup into `makeCallWithMockCoordinator()` and reused it in that test too, which removed four duplicated lines.

### 🧪 Manual Testing Notes

```swift
let call = streamVideo.call(callType: "default", callId: callId)
try await call.create()

// Soft delete
try await call.delete()

// Hard delete — removes recordings, transcriptions and session history.
// `taskId` identifies the async backend task.
let response = try await call.delete(hard: true)
print(response.taskId ?? "-")
```

A subsequent `call.get()` should fail, and other members should receive a `CallDeletedEvent`.

### ☑️ Contributor Checklist

- [ ] Affected documentation updated (tutorial, CMS) — the iOS call-lifecycle docs page should gain a `delete` mention; that lives outside this repo

## Summary by CodeRabbit

- **New Features**
  - Added asynchronous call deletion.
  - Supports soft deletion by default and optional hard deletion of related call data.
  - Returns deletion results and reports errors when the operation fails.

- **Bug Fixes**
  - Call state now updates correctly when a call is deleted.

- **Documentation**
  - Added call deletion examples to the call creation and joining guides.
  - Added an upcoming changelog entry covering call deletion and hard-delete behavior.

### DIFF
--- a/Sources/StreamVideo/Call.swift
+++ b/Sources/StreamVideo/Call.swift
@@ -831,6 +831,28 @@ public class Call: @unchecked Sendable, WSEventsSubscriber {
         try await coordinatorClient.endCall(type: callType, id: callId)
     }
 
+    /// Deletes the call.
+    ///
+    /// Unlike ``end()``, which only stops the active session, deleting removes
+    /// the call itself. The backend emits a `CallDeletedEvent` to the remaining
+    /// members. Participants that are still in the call are not detached
+    /// automatically, so call ``leave(reason:)`` if the current user has joined.
+    ///
+    /// - Parameter hard: When `true`, the call is hard deleted along with all
+    ///   related data (recordings, transcriptions, session history). Hard
+    ///   deletion is performed asynchronously on the backend, in which case
+    ///   `DeleteCallResponse.taskId` identifies the task. Defaults to `false`.
+    /// - Returns: A `DeleteCallResponse` describing the deleted call.
+    /// - Throws: An error if deleting the call fails.
+    @discardableResult
+    public func delete(hard: Bool = false) async throws -> DeleteCallResponse {
+        try await coordinatorClient.deleteCall(
+            type: callType,
+            id: callId,
+            deleteCallRequest: DeleteCallRequest(hard: hard)
+        )
+    }
+
     /// Blocks a user in a call.
     /// - Parameters:
     ///   - userId: The ID of the user to block.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ### ✅ Added
-- Added `Call.delete(hard:)` for deleting a call. Pass `hard: true` to also remove all related data.
+- Added `Call.delete(hard:)` for deleting a call. Pass `hard: true` to also remove all related data. [#1243](https://github.com/GetStream/stream-video-swift/pull/1243)
 
 ### 🐞 Fixed
 - Transient peer-connection disconnections no longer trigger an immediate full rejoin, allowing the existing ICE restart flow to recover the session. [#1231](https://github.com/GetStream/stream-video-swift/pull/1231)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ### ✅ Added
-- Added `Call.delete(hard:)` for deleting a call. Pass `hard: true` to also remove all related data. [#1243](https://github.com/GetStream/stream-video-swift/pull/1243)
+- Added `Call.delete(hard:)` for deleting a call. [#1243](https://github.com/GetStream/stream-video-swift/pull/1243)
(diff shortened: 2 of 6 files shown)


martinmitrevski:
Added a second change after review feedback: `CallDeletedEvent` now updates `Call.state` instead of being discarded.

**What JS does.** In the whole `stream-video-js` monorepo, `call.deleted` is handled in exactly one place:

```ts
// packages/client/src/store/CallState.ts:515
'call.deleted': (e) => this.updateFromCallResponse(e.call),
```

That is the entire handling. Nothing in `react-sdk`, nothing in `react-native-sdk`, no `callingX` teardown, and no leave. The Swift equivalent of `updateFromCallResponse(call)` is `update(from: event.call)`, and that is what this now does, sitting right next to the identical `CallEndedEvent` case.

**Why it does not leave the call.** Worth recording, since it came up:

- No SDK leaves on delete. JS updates state only; Android does not handle `CallDeletedEvent` at all.
- On iOS, joined-participant teardown runs off the *SFU* `CallEnded` event (`WebRTCCoordinator+Joined.observeCallEndedEvent`), not a coordinator event. JS instead leaves off the *coordinator* `call.ended` event (`watchCallEnded`). The two platforms already differ architecturally here, so adding a coordinator-driven leave on iOS alone would create a new divergence rather than remove one.
- Whether a joined participant is torn down on delete depends on whether the backend emits an SFU `CallEnded`. `Stream_Video_Sfu_Models_CallEndedReason` has only `unspecified / ended / liveEnded / kicked / sessionEnded`, no `deleted`, which suggests it does not, but that is inference from the client side and worth confirming with the backend team.

**One gap left open, deliberately.** If a call is deleted while it is *ringing*, CallKit keeps ringing. `CallKitService`'s event subscription handles `CallEndedEvent`, `CallAcceptedEvent`, `CallRejectedEvent` and `CallSessionParticipantLeftEvent` and `default: break`s the rest, and the `checkIfCallWasHandled` gate only runs once at ring start. A ringing device has not joined the SFU, so no SFU event can reach it either. This one is iOS-specific with no JS counterpart, so it is not a parity question. Happy to fix it here or in a follow-up, whichever you prefer.

Tests: `test_updateState_fromCallDeletedEvent` asserts the response lands in state (`endedAt`), and fails if the handler is reverted to `break`. `Call_Tests` and `CallState_Tests` are green (62 cases).

Sources/StreamVideo/Call.swift
```diff
@@ -831,6 +831,28 @@ public class Call: @unchecked Sendable, WSEventsSubscriber {
...
+    /// automatically, so call ``leave(reason:)`` if the current user has joined.
+    ///
+    /// - Parameter hard: When `true`, the call is hard deleted along with all
+    ///   related data (recordings, transcriptions, session history). Hard
+    ///   deletion is performed asynchronously on the backend, in which case
+    ///   `DeleteCallResponse.taskId` identifies the task. Defaults to `false`.
+    /// - Returns: A `DeleteCallResponse` describing the deleted call.
+    /// - Throws: An error if deleting the call fails.
+    @discardableResult
+    public func delete(hard: Bool = false) async throws -> DeleteCallResponse {
```
ipavlidakis, line 848: TBH it seems weird to me that we aren't leaving. Especially during a hard delete, the session is being deleted which means that probably the SFU will also drop all connections. Did you actually try it? what is the behaviour? Do we receive any additional events from the SFU that maybe those can be responsible for kicking the user out?

martinmitrevski: checked this, the SFU kicks you out of the call

ipavlidakis: Can we update the comment with this information/expectation, please?

ipavlidakis (approved):

PR #1246, opened 2026-08-17
Expose frame recording on Call

martinmitrevski:
## 🎯 Goal

Frame recording was reachable on the JS SDK but not on iOS. This closes that gap (audit finding **G2**, and the `Frame recording` row of the capability matrix).

## 🛠 Implementation

`startFrameRecording` / `stopFrameRecording` were already generated on the concrete `DefaultAPI`, but they were the **only two** endpoints missing from the `DefaultAPIEndpoints` protocol. `Call.coordinatorClient` is typed as that protocol, so frame recording was genuinely unreachable rather than merely un-wrapped — a wrapper alone would not have compiled. I checked the other nine endpoints the audit lists as unexposed (`deleteCall`, `getEdges`, `listTranscriptions`, …) and all nine are already in the protocol, so this is specific to frame recording.

- **`Call`** — `startFrameRecording(recordingExternalStorage:)` and `stopFrameRecording()`.
- **`CallState`** — new `frameRecordingStatus`. The four frame recording events were all matched and discarded (`break`); started sets `true`, stopped/failed set `false`, and `frameReady` stays a no-op since captured frames are delivered through event subscriptions. Also derived from `egress.frameRecording` on every call response, using the same `"running"` rule already applied to the composite, individual and raw recording flags.
- **`DefaultAPI.swift`** — the two missing protocol requirements, placed where the generator's own ordering puts them. This file is excluded from swiftformat (`--exclude **/OpenApi`), so the existing separator whitespace is preserved to keep regeneration a no-op.

## 🤝 Consistency with the other SDKs

JS is the only other platform that exposes this, and it applies **no capability check** — `OwnCapability` has no frame recording case, so none was added here either. Swift's convention is to flatten a single-field request object into a named parameter, exactly as `startTranscription(transcriptionExternalStorage:)` does, so `recordingExternalStorage: String? = nil` is the matching shape rather than passing a request object.

Neither SDK previously tracked frame recording status. JS's `updateFromCallResponse` derives `recording` / `individualRecording` / `rawRecording` with `status === 'running'` — the identical rule Swift's `didUpdate(_ egress:)` already used — so `frameRecordingStatus` extends that existing family rather than introducing a new pattern.

Mode, quality and capture interval are configured through the call type's frame recording settings and were already reachable via `Call.update(settingsOverride:)` (`CallSettingsRequest.frameRecording`), so they were never part of this gap.

## 🧪 Testing

Nine new tests, all passing (`StreamVideoTests`, iPhone 16 / iOS 18.3.1):

- coordinator called with the expected type, id and payload for start (with and without external storage) and stop
- `frameRecordingStatus` transitions for the started, stopped and failed events, and that `frameReady` leaves it unchanged
- `frameRecordingStatus` set from a call response carrying a running frame recording, and reset when it is absent

⚠️ One thing worth flagging for CI: the first test executed in `Call_Tests` absorbs the process cold-start cost and timed out against XCTest's 2-minute per-test allowance on the first cold run here. It passes consistently once the build is warm (verified by running it behind an earlier-sorting test). This is a pre-existing property of the suite, not something this PR introduces.

## ☑️ Contributor Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform)
- [x] Documented code (public API documented with DocC comments)
- [x] Changelog updated
- [x] Unit tests added

### DIFF
--- a/Sources/StreamVideo/OpenApi/generated/APIs/DefaultAPI.swift
+++ b/Sources/StreamVideo/OpenApi/generated/APIs/DefaultAPI.swift
@@ -1281,6 +1281,9 @@ protocol DefaultAPIEndpoints {
     func startClosedCaptions(type: String, id: String, startClosedCaptionsRequest: StartClosedCaptionsRequest) async throws
         -> StartClosedCaptionsResponse
         
+    func startFrameRecording(type: String, id: String, startFrameRecordingRequest: StartFrameRecordingRequest) async throws
+        -> StartFrameRecordingResponse
+        
     func startRecording(
         type: String,
         id: String,
@@ -1295,6 +1298,8 @@ protocol DefaultAPIEndpoints {
         
     func stopClosedCaptions(type: String, id: String, stopClosedCaptionsRequest: StopClosedCaptionsRequest) async throws
         -> StopClosedCaptionsResponse
+        
+    func stopFrameRecording(type: String, id: String) async throws -> StopFrameRecordingResponse
         
     func stopLive(type: String, id: String, stopLiveRequest: StopLiveRequest) async throws -> StopLiveResponse
         
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### ✅ Added
+- `Call.startFrameRecording(recordingExternalStorage:)` and `Call.stopFrameRecording()`, along with the `CallState.frameRecordingStatus` property that tracks frame recording for the active call.
+
 ### 🐞 Fixed
 - Transient peer-connection disconnections no longer trigger an immediate full rejoin, allowing the existing ICE restart flow to recover the session. [#1231](https://github.com/GetStream/stream-video-swift/pull/1231)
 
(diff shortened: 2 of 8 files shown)


Sources/StreamVideo/OpenApi/generated/APIs/DefaultAPI.swift
```diff
@@ -1281,6 +1281,9 @@ protocol DefaultAPIEndpoints {
     func startClosedCaptions(type: String, id: String, startClosedCaptionsRequest: StartClosedCaptionsRequest) async throws
         -> StartClosedCaptionsResponse
         
+    func startFrameRecording(type: String, id: String, startFrameRecordingRequest: StartFrameRecordingRequest) async throws
```
ipavlidakis, line 1284: Is that generated or manually added?

ipavlidakis (approved):
I added a question that requires clarification but mostly looks good?

Sources/StreamVideo/OpenApi/generated/APIs/DefaultAPI.swift
martinmitrevski: it's generated, for some reason it was removed until now (the implementation was there)

PR #1249, opened 2026-08-18
[Enhancement] Add framed AES-GCM E2EE for calls

ipavlidakis:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1970/e2ee-calls-on-video

Depends on https://github.com/GetStream/webrtc/pull/110

### 🎯 Goal

Let iOS join E2EE calls with the same framed AES-GCM scheme as the JS SDK: attach keys before join, encrypt local senders, decrypt remote receivers, and report `e2ee: true` on `JoinCallRequest`.

### 📝 Summary

- Add `E2EEManager` / `EncryptionManager` and `Call.setE2EEManager(_:)`. Must run before join; later adoption throws so a live session cannot go half-encrypted.
- Join sends `JoinCallRequest.e2ee` when a manager is set. The manager survives `leave` so a later join still encrypts.
- Encrypt at `addTransceiver` via `E2EEAttachmentContext`. Decrypt when subscriber tracks arrive; queue until the remote `userId` is known.
- Match RTP receivers by `trackId`. `RTCRtpReceiver.track` allocates a new ObjC wrapper each read, so identity (`===`) never attached decryptors (distorted video, silent audio).
- Demo debug menu accepts 32/64-char hex or a web-style PBKDF2 passphrase (`stream-e2ee`, 100k iterations, 16-byte AES-128) and attaches keys before join.
- Log native `e2ee.*` events on `.webRTC`.
- Unit tests for manager construction/keys, call/controller attach, and decryptor attach (including unknown participant then flush).

### 🛠 Implementation

- `EncryptionManager` wraps `RTCEncryptionManager`: AES-128/256, shared and per-user keys, codec pins (`opus` / `vp8` / `vp9` / `h264`), replay grouped by `TrackType`.
- Publisher and subscriber coordinators each own an `E2EEAttachmentContext`. Local audio/video/screenshare adapters encrypt when they add a transceiver.
- `WebRTCStateAdapter` attaches decryptors from subscriber track events, using `rtpReceiver(matching:)` (`trackId`), and retries `pendingDecryptors` after participant updates.
- Demo `AppEnvironment.EncryptionKeys` is the only key UI. `attachIfNeeded` is called from lobby/join paths so encryption is on before peer connections exist.

### 🎨 Showcase

Not applicable. Debug-menu key entry is text-only (hex or passphrase sheet).

|  Before  |  After  |
| -------- | ------- |
|  `img`   |  `img`  |

### 🧪 Manual Testing Notes

1. Rebuild/point the demo at a StreamWebRTC that includes `RTCEncryptionManager`.
2. Debug menu → Encryption Key → paste the same passphrase or hex as web (e.g. `noun-rover-waitress`).
3. Join an E2EE call with a JS client using that key.
4. Confirm join reports `e2ee: true`, logs show encrypt/decrypt attach, and video is clear.
5. Audio both directions still needs the WebRTC audio-receiver SSRC attach fix in StreamWebRTC; without that rebuild, encrypted audio can stay silent even when Swift attached the decryptor.

### ☑️ Contributor Checklist

### 🎁 Meme

## Summary by CodeRabbit

* **New Features**
  * Added AES-GCM end-to-end encryption for audio, video, and screen sharing.
  * Added shared and per-user key management with passphrase and hex key support.
  * Added encryption configuration to call creation, joining, ringing, lobby, and deeplink flows.
  * Added encryption status indicators and key controls in the demo interface.
  * Added support for clearing encryption settings and handling encryption keys from invitation links and QR codes.

* **Documentation**
  * Updated encryption configuration and call setup documentation.

* **Tests**
  * Added coverage for key validation, media encryption, call setup, and delayed decryption.


DemoApp/Sources/Components/Debug/Items/FeatureFlags/Components/EncryptionKey.swift
```diff
@@ -0,0 +1,329 @@
...
+            prompt = nil
+        }
+
+        /// Asks the root debug menu to show the hex editor after UIMenu dismisses.
+        func request(_ prompt: Prompt) {
+            Task { [weak self] in
+                try await Task.sleep(nanoseconds: 350_000_000)
+                self?.prompt = prompt
+            }
+        }
```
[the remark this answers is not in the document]
ipavlidakis: Not addressing. This is DemoApp debug-menu UI, not SDK code. The prompt is presented from SwiftUI, and a thrown `Task.sleep` on cancel is harmless here.

DemoApp/Sources/Components/Debug/Items/FeatureFlags/Components/EncryptionKey.swift
```diff
@@ -0,0 +1,329 @@
...
+                }
+                try await call.setE2EEManager(e2ee)
+                log.debug(
+                    "Demo E2EE attached with \(keys.count) shared key(s).",
+                    subsystems: .webRTC
+                )
+            } catch {
+                log.error(error, subsystems: .webRTC)
+            }
+        }
```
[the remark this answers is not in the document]
ipavlidakis: Not addressing in the SDK. `setE2EEManager` already throws after peer connections exist, and `attachIfNeeded` logs that. Tracking `cId` would only paper over DemoApp calling attach from several entry points.

Sources/StreamVideo/Call.swift
```diff
@@ -583,6 +588,39 @@ public class Call: @unchecked Sendable, WSEventsSubscriber {
         callController.setVideoFilter(videoFilter)
     }
 
+    /// The E2EE manager for this call, if one was set via ``setE2EEManager(_:)``.
+    ///
+    /// Kept across ``leave(reason:)`` so a later join still reports `e2ee: true`
+    /// and still attaches encrypt/decrypt on the new peer connections.
+    public private(set) var e2eeManager: E2EEManager?
```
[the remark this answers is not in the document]
ipavlidakis: Addressed by removing those copies. `Call` no longer stores a manager, and `CallController` no longer keeps a second one. The session owner is `WebRTCStateAdapter` (an actor), which join reads for `JoinCallRequest.e2ee`.

Sources/StreamVideo/Call.swift
```diff
@@ -583,6 +588,39 @@ public class Call: @unchecked Sendable, WSEventsSubscriber {
...
+    ///
+    /// - Parameter manager: The encryption manager to attach.
+    /// - Throws: If called after peer connections exist (i.e. after join).
+    ///   Those PCs were already configured without an encryptor, so adopting
+    ///   a manager now would silently publish or receive cleartext.
+    public func setE2EEManager(_ manager: E2EEManager) async throws {
+        try await callController.setE2EEManager(manager)
+        e2eeManager = manager
+    }
+
```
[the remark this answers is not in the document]
ipavlidakis: Addressing. Added an Upcoming changelog entry for `Call.setE2EEManager(_:)` and `EncryptionManager`. No migration notes — this is a new API, not a replacement of an existing one.

Sources/StreamVideo/WebRTC/E2EE/E2EEAttachmentContext.swift
```diff
@@ -0,0 +1,57 @@
...
+        guard let manager else { return }
+        do {
+            try manager.encrypt(sender, codec: codec, trackType: trackType)
+            log.debug(
+                "E2EE encryptor attached to sender codec:\(codec ?? "nil") trackType:\(trackType).",
+                subsystems: .webRTC
+            )
+        } catch {
+            log.error(error, subsystems: .webRTC)
+        }
```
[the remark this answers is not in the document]
ipavlidakis: Not addressing. Fail-open is intentional and matches the JS SDK: a missing key or encrypt failure must not abort transceiver creation. Failures are logged on `encryptIfNeeded`.

Sources/StreamVideo/WebRTC/E2EE/E2EEAttachmentContext.swift
```diff
@@ -0,0 +1,57 @@
...
+    func encryptIfNeeded(
+        sender: RTCRtpSender,
+        codec: String?,
+        trackType: TrackType
+    ) {
+        guard let manager else { return }
+        do {
+            try manager.encrypt(sender, codec: codec, trackType: trackType)
+            log.debug(
+                "E2EE encryptor attached to sender codec:\(codec ?? "nil") trackType:\(trackType).",
```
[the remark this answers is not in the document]
ipavlidakis: Not addressing. Wrapping these log strings to 80 characters would not change behavior.

Sources/StreamVideo/WebRTC/v2/PeerConnection/MediaAdapters/LocalMediaAdapters/LocalAudioMediaAdapter.swift
```diff
@@ -366,6 +375,11 @@ final class LocalAudioMediaAdapter: LocalMediaAdapting, @unchecked Sendable {
             return
         }
         transceiverStorage.set(transceiver, track: track, for: options)
+        e2ee.encryptIfNeeded(
+            sender: transceiver.sender,
+            codec: options.codec.e2eeCodecPin,
+            trackType: .audio
+        )
```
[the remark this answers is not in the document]
ipavlidakis: Not addressing. Same as `encryptIfNeeded`: fail-open is the product contract. Callers must not drop the transceiver and publish nothing when encrypt attachment fails.

StreamVideoTests/WebRTC/v2/PeerConnection/MediaAdapters/LocalMediaAdapters/LocalAudioMediaAdapter_Tests.swift
```diff
@@ -293,6 +296,24 @@ final class LocalAudioMediaAdapter_Tests: XCTestCase, @unchecked Sendable {
...
+                with: try makeTransceiver(of: .audio, audioOptions: publishOption)
+            )
+        }
+
+        try await subject.publish()
+
+        await fulfillment { mockE2EE.encryptCalls.count == 1 }
+        XCTAssertEqual(mockE2EE.encryptCalls.first?.codec, "opus")
+        XCTAssertEqual(mockE2EE.encryptCalls.first?.trackType, .audio)
+    }
```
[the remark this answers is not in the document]
ipavlidakis: Not addressing. The current names already describe condition, action, and result. Renaming tests is out of scope for this PR.

DemoApp/Sources/Components/Debug/Items/FeatureFlags/Components/EncryptionKey.swift
```diff
@@ -0,0 +1,407 @@
...
+                throw EncryptionKeyError.mixedLength
+            }
+            guard length == 16 || length == 32 else {
+                throw EncryptionKeyError.wrongLength(length)
+            }
+        }
+    }
+}
+
+extension DebugMenu {
```
martinmitrevski, line 172: shouldn't we move the view code in a separate file? I assume people can just copy-paste this one (and modify some values).

DemoApp/Sources/Components/DemoAppViewFactory.swift
```diff
@@ -22,15 +23,12 @@ final class DemoAppViewFactory: ViewFactory {
...
-        DefaultViewFactory
-            .shared
-            .makeLobbyView(
-                viewModel: viewModel,
-                lobbyInfo: lobbyInfo,
-                callSettings: callSettings
-            )
-            .alignedToReadableContentGuide()
-            .background(Appearance.default.colors.lobbyBackground.edgesIgnoringSafeArea(.all))
+        DemoLobbyView(
```
martinmitrevski: not a fan of this tbh. We're re-implementing the whole view - which is also tricky for testing, but it also makes it a bit harder for customers to add e2ee. Can't we expose some part of the view to inject this feature on the default view?

DemoApp/Sources/Views/CallTopView/DemoCallTopView.swift
```diff
@@ -44,7 +44,10 @@ struct DemoCallTopView<Factory: ViewFactory>: View {
 
             if !isCallLivestream {
                 HStack(alignment: .center) {
-                    CallDurationView(viewModel)
+                    DemoCallDurationView(
```
martinmitrevski, line 47: Similarly here - the demo app should use the default components as much as possible

DemoApp/Sources/Views/CallTopView/DemoCallTopView.swift
```diff
@@ -140,6 +147,85 @@ struct DemoCallTopView<Factory: ViewFactory>: View {
     }
 }
 
+private struct DemoCallDurationView: View {
```
martinmitrevski, line 150: if it makes sense, we should expose a way to add the view indicator, but I'm also fine with the default implementation.

DemoApp/Sources/Views/CallView/CallingView/DemoLobbyEncryptionSection.swift
```diff
@@ -0,0 +1,377 @@
...
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+import StreamVideo
+import StreamVideoSwiftUI
+import SwiftUI
+
+/// Demo `ViewFactory.makeLobbyView` override. Same lobby as the SDK, with an
+/// extra encryption control on the call-settings row.
+struct DemoLobbyView: View {
```
martinmitrevski: Ideally all this goes away and we expose some slot in the lobby view

Sources/StreamVideo/WebRTC/E2EE/E2EEAttachmentContext.swift
```diff
@@ -0,0 +1,57 @@
...
+/// Publisher peer connections create audio, video, and screenshare adapters
+/// from a single context. ``RTCPeerConnectionCoordinator/attachE2EE(_:)``
+/// writes the manager after construction and before `setUp`, so
+/// `addTransceiver` can encrypt without exploding every initializer
+/// signature.
+///
+/// Access is serialized on `UnfairQueue` because local adapters publish
+/// from WebRTC callbacks while the coordinator attaches the manager from
+/// the state-adapter actor.
+final class E2EEAttachmentContext: @unchecked Sendable {
```
martinmitrevski, line 21: Why do we swallow the errors here? Maybe good to document the reasoning here.

Sources/StreamVideo/WebRTC/E2EE/EncryptionManager.swift
```diff
@@ -0,0 +1,376 @@
...
+
+import Combine
+import Foundation
+import StreamWebRTC
+
+/// AES-GCM variant used by ``EncryptionManager``.
+///
+/// - `'aes128Gcm'` (default): 16-byte keys.
+/// - `'aes256Gcm'`: 32-byte keys.
+public enum E2EEAlgorithm: Sendable {
```
martinmitrevski, line 13: is an enum the best option here? If we need to add another algorithm, that would be a breaking change.

Sources/StreamVideo/WebRTC/E2EE/EncryptionManager.swift
```diff
@@ -0,0 +1,376 @@
...
+            log.warning(message, subsystems: .webRTC)
+        } else {
+            log.debug(message, subsystems: .webRTC)
+        }
+        handler(mapped)
+    }
+}
+
+/// Native already throttles unencrypted_frame and unsupported_version.
+func e2eeEventLogsAsWarning(_ type: RTCE2eeEventType) -> Bool {
```
martinmitrevski: why not part of the class?

Sources/StreamVideo/Call.swift
```diff
@@ -583,6 +592,36 @@ public class Call: @unchecked Sendable, WSEventsSubscriber {
         callController.setVideoFilter(videoFilter)
     }
 
+    /// Attaches or clears end-to-end encryption. Must run before ``join``.
+    ///
+    /// ## Overview
+    /// Pass ``EncryptionManager`` for the built-in AES-GCM scheme, or any
+    /// custom ``E2EEManager``. The manager is stored on the WebRTC session
+    /// so join:
+    /// - reports `e2ee: true` on ``JoinCallRequest``
```
martinmitrevski: some of these are too verbose for a public api

martinmitrevski (commented):
Imo, we should avoid replacing the standard components - it will make our live harder for testing and maintenance.

DemoApp/Sources/Components/DemoAppViewFactory.swift
ipavlidakis: I agree that re-implemnting isn't the best option but ... it was the only one without changing our Public API. Ofc we can expose something new but adding slots like that requires a little bit more thinking than it took me to implement this

DemoApp/Sources/Views/CallTopView/DemoCallTopView.swift
ipavlidakis: That will then require quite (if not a lot) changes on the public api and the slots as the things we need to change are "surgical" inside our existing view slots.

Sources/StreamVideoSwiftUI/CallingViews/PreJoiningView.swift
```diff
@@ -72,6 +76,7 @@ struct LobbyContentView<Factory: ViewFactory>: View {
     var callId: String
     var callType: String
     @Binding var callSettings: CallSettings
+    var callSettingsView: (Binding<CallSettings>) -> SettingsView
```
martinmitrevski, line 79: isn't it better to expose a slot from the factory?

DemoApp/Sources/Views/CallTopView/DemoCallTopView.swift
martinmitrevski: I don't see much value in this one - best if we just use the default view. If customers want to show some indicator for that, they should build their own ui.

martinmitrevski (commented):
left few comments, plus the ones from before are still not addressed

DemoApp/Sources/Views/CallTopView/DemoCallTopView.swift
ipavlidakis: RN, Flutter and React have updated their Demo UI to show this one. Happy to drop it but we will be different from the rest.

Sources/StreamVideoSwiftUI/CallingViews/PreJoiningView.swift
ipavlidakis: No, this is exactly what i would like to avoid. The ViewFactory should contain high level ViewSlots not every slot for every View that we may want to make configurable as this will increase the responsibilities and size of the ViewFactory (which is supposed to be generic).

Sources/StreamVideo/WebRTC/E2EE/EncryptionManager.swift
ipavlidakis: Happy to change it if you prefer something more open-ended, but I don’t see much value right now. The closed enum matches the algorithms we actually have: AES-128-GCM and AES-256-GCM (same as JS / native). That’s two key sizes of the same construction, not a list we expect to keep extending — a new algorithm would need a native change and a new framing version, so it would be a minor bump anyway, and we don’t expect that often.

Adding a case to a public enum is only source-breaking if someone `switch`es over every case. That’s not a realistic use of this type — callers pass one of the two values into `EncryptionManager`, they don’t switch on it.

DemoApp/Sources/Components/Debug/Items/FeatureFlags/Components/EncryptionKey.swift
```diff
@@ -0,0 +1,407 @@
...
+
+        var wantsEncryption: Bool {
+            !values.isEmpty || !shareKey.isEmpty
+        }
+
+        var shareKey: String {
+            let trimmed = passphrase.trimmingCharacters(in: .whitespacesAndNewlines)
+            if !trimmed.isEmpty { return trimmed }
+            return values.first.map(DemoHex.string(from:)) ?? ""
+        }
```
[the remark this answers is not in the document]
ipavlidakis: Fixed: `shareKey` now returns empty when there isn't exactly one key, so many-keys mode omits the invite query parameter instead of sharing only key 0.

DemoApp/Sources/Components/Deeplinks/DeeplinkAdapter.swift
```diff
@@ -92,10 +95,49 @@ struct DeeplinkAdapter {
...
+            let end = rest.firstIndex {
+                $0 == "&" || $0 == "#" || $0.isNewline || $0 == " "
+            } ?? rest.endIndex
+            var value = String(rest[..<end])
+            value = value.removingPercentEncoding ?? value
+            let trimmed = value.trimmingCharacters(in: .whitespacesAndNewlines)
+            if !trimmed.isEmpty { return trimmed }
+        }
+        return nil
+    }
```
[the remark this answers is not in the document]
ipavlidakis: Fixed: both the query-item and raw parsers now form-decode `+` as a space before percent-decoding, so web passphrases like `able+acid+anger` match PBKDF2.

DemoApp/Sources/Views/CallView/CallingView/DemoLobbyEncryptionSection.swift
```diff
@@ -0,0 +1,377 @@
...
+                            .accessibility(identifier: "e2eeKey")
+
+                        Button(action: keys.generatePassphrase) {
+                            Image(systemName: "arrow.triangle.2.circlepath")
+                                .foregroundColor(
+                                    Color(colors.textLowEmphasis)
+                                )
+                        }
+                        .buttonStyle(BorderlessButtonStyle())
+                        .accessibility(identifier: "e2eeNewKey")
```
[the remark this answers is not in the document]
ipavlidakis: Fixed: the regenerate-key button now has `accessibilityLabel("Generate a new shared key")`. The control lives in `DemoCallSettingsView` after the file move.

DemoApp/Sources/Views/CallView/CallingView/SimpleCallingView.swift
```diff
@@ -156,6 +156,8 @@ struct SimpleCallingView: View {
             return
         }
 
+        AppEnvironment.EncryptionKeys.shared.applyDeeplink(deeplinkInfo)
+
```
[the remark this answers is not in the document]
ipavlidakis: Fixed: `applyDeeplink` now runs only in the branch that accepts the link, after the environment check. Declining a different-environment prompt no longer keeps that key.

Sources/StreamVideo/Call.swift
ipavlidakis: Trimmed it. `setE2EEManager` now matches `setVideoFilter:` attach or clear before join, nil detaches, throws after peer connections exist.

Sources/StreamVideo/WebRTC/E2EE/EncryptionManager.swift
ipavlidakis: Moved it off `EncryptionManager`. The switch is on `RTCE2eeEventType`, so it’s a private logsAsWarning on that type in the same file. The forwarder uses event.type.logsAsWarning; nothing else needs to see it.

DemoApp/Sources/Views/CallView/CallingView/DemoLobbyEncryptionSection.swift
ipavlidakis: This is already covered: LobbyView takes callSettingsView. The demo keeps the default lobby and injects DemoCallSettingsView there. We did not add a ViewFactory method on purpose — factory stays high-level.

Sources/StreamVideo/WebRTC/E2EE/E2EEAttachmentContext.swift
ipavlidakis: Documented on encryptIfNeeded: we log and don’t rethrow so a missing key or disposed manager can’t abort transceiver creation. Same fail-open as JS.

DemoApp/Sources/Components/Debug/Items/FeatureFlags/Components/EncryptionKey.swift
ipavlidakis: The copy-paste sample is DemoCallSettingsView (lobby). EncryptionKey.swift is debug-menu plumbing (hex editor / many-keys), not the snippet I’d want people to copy.

DemoApp/Sources/Views/CallTopView/DemoCallTopView.swift
martinmitrevski: ok, I guess it's not that big of a deal to have this for consistency with the other sdks

Sources/StreamVideo/WebRTC/E2EE/EncryptionManager.swift
martinmitrevski: yeah, typically we do static lets in these cases. Your argument that maybe people will not do switch over it also makes sense, so I'm leaving this one up to you.

Sources/StreamVideoSwiftUI/CallingViews/PreJoiningView.swift
martinmitrevski: fair enough

Sources/StreamVideo/WebRTC/E2EE/EncryptionManager.swift
```diff
@@ -0,0 +1,377 @@
...
+        }
+    }
+}
+
+/// One `e2ee.*` event from the native encryption manager.
+///
+/// Events include missing keys, decryption failures, and optional
+/// performance reports. Optional fields are `nil` when the native event
+/// omitted them.
+public struct E2EEEvent: Sendable {
```
martinmitrevski, line 35: maybe final class would help with size here?

martinmitrevski (approved):
LGTM ✅ Left few small comments, but nothing really blocking

Sources/StreamVideo/WebRTC/E2EE/EncryptionManager.swift
ipavlidakis: `EncryptionManager` is already `public final class`. This type is `E2EEEvent`, a small all-`let` snapshot we publish on Combine.

A `final class` here would heap-allocate on every native callback and change public value semantics. The +12KB in `EncryptionManager.o` is the manager + native bridge, not this payload. I’d keep the struct unless you want it measured.

PR #1250, opened 2026-08-18
Apply SFU call grants on own capabilities

martinmitrevski:
### 🔗 Issue Links

N/A — found while auditing behaviour parity against the JavaScript and Android SDKs.

### 🎯 Goal

The SFU sends `callGrantsUpdated` when the local participant's publishing rights change during a call. We never subscribed to it: the protobuf type appears in the SDK only in the debug-description helpers.

So when a moderator revokes audio, video or screenshare mid-call, `Call.state.ownCapabilities` keeps reporting the old value and the UI controls stay enabled until an unrelated coordinator event happens to arrive. The JavaScript SDK applies these grants; both mobile SDKs do not.

### 📝 Summary

- New internal `CallGrants` model that maps the SFU payload and knows how to apply itself to a capability list.
- `SFUEventAdapter` subscribes to `callGrantsUpdated` and stores the grants on `WebRTCStateAdapter`.
- `CallController` bridges them up to `Call.state`.
- `CallState` re-applies the latest grants whenever the coordinator reports a new capability set, so grants take precedence.
- 12 unit tests across the four layers. Changelog updated.

### 🛠 Implementation

Grants enter at the SFU boundary and travel up the path the codebase already uses for `anonymousCount` and `isSpeakingWhileMuted`, rather than reaching into the call state from the WebRTC layer:

```
SFU callGrantsUpdated
  → SFUEventAdapter.handleCallGrantsUpdated      (guarded on event.hasCurrentGrants)
  → WebRTCStateAdapter.callGrants                (@Published, reset in cleanUp)
  → CallController.observeCallGrantsUpdates()    (shaped like observeSpeakingWhileMutedUpdates)
  → CallState.update(callGrants:)                (merges into ownCapabilities)
```

`CallGrants.applied(to:)` adds a granted capability when it's missing and removes a revoked one when it's present, leaving every other capability untouched. It is idempotent, so re-applying the same grants is safe. `CallState` keeps the last grants and runs them over each new coordinator capability set, which is what makes grants win — the five `ownCapabilities = response.ownCapabilities` assignments and the `UpdatedCallPermissionsEvent` path now route through one private `setOwnCapabilities`.

Three decisions worth a reviewer's attention:

1. **Grants persist across reconnection.** They are cleared in `cleanUp` (end of call) but deliberately not in `cleanUpForReconnection`, which already preserves `ownCapabilities` and keeps `initialCallSettings` "in order to maintain the state during reconnects". Clearing them there would silently restore revoked rights on a fast reconnect, because the SFU only re-sends grants when they change.

2. **Revoking a capability also mutes the matching track**, and this comes for free rather than being added here: `Call+JoinedStage` already observes `state.$ownCapabilities` and pushes changes into `enqueueOwnCapabilities`, which turns off audio/video and calls `stopScreenSharing`. No feedback loop, since `callGrants` and `ownCapabilities` are separate properties on the adapter and `updateOwnCapabilities` guards on inequality.

3. **`CallGrants` is internal**, so the public API surface is unchanged. JS does not expose grants either, only the merged `ownCapabilities`. Happy to make it public if you'd rather apps could read the raw grants.

### 🧪 Manual Testing Notes

With two clients in a call, from the second client (or the dashboard/server) revoke the first client's `send-audio` permission. On `develop` the first client's `state.ownCapabilities` still contains `.sendAudio` and its mic button stays enabled; with this change the capability disappears and the audio track is muted. Granting it back restores both.

### ☑️ Contributor Checklist

- [ ] Affected documentation updated (tutorial, CMS) — no docs change needed; this makes an already-documented property behave as documented

## Summary by CodeRabbit

* **Bug Fixes**
  * Mid-call permission changes now update available call capabilities immediately.
  * Revoking audio, video, or screen-sharing permissions correctly disables the affected activity.
  * Call capabilities remain synchronized with the current call state.
  * Repeated or incomplete permission updates are handled safely without disrupting existing capabilities.
  * Reloading a ringing call after a WebSocket reconnect now works correctly.

* **Changed**
  * Screensharing participants now rank above pinned participants in default speaker layouts.


Sources/StreamVideo/WebRTC/v2/SFU/SFUEventAdapter.swift
```diff
@@ -556,4 +564,23 @@ final class SFUEventAdapter: @unchecked Sendable {
...
+    /// reported by the coordinator, so they are stored and later merged into
+    /// `Call.state.ownCapabilities`.
+    private func handleCallGrantsUpdated(
+        _ event: Stream_Video_Sfu_Event_CallGrantsUpdated
+    ) async {
+        guard event.hasCurrentGrants else {
+            return
+        }
+
+        await stateAdapter.set(callGrants: .init(event.currentGrants))
```
ipavlidakis: I'm not sure about this one. CallGrants seems to be another wrapper around OwnCapabiliities. I would prefer to not "pollute" the StateAdapter with a passthrough property and instead, here convert to OwnCapablities and enqueue the change. Then in the CallController you can create a new subscription that will observe the OwnCapabilities from the StateAdapter and will pass them through the Call.state.

martinmitrevski: Agreed, and done in eb394e3e. `CallGrants` and the state-adapter property are both gone.

`handleCallGrantsUpdated` now converts and enqueues:

```swift
let grants = event.currentGrants
await stateAdapter.enqueueOwnCapabilities { grants.applied(to: $0) }
```

`applied(to:)` lives in `SFU/Extensions/Stream_Video_Sfu_Models_CallGrants+OwnCapability.swift`, so the mapping stays next to the other SFU extensions and there is no new wrapper type. `CallController` got `observeOwnCapabilitiesUpdates()` observing `stateAdapter.$ownCapabilities` and forwarding to `Call.state`, and `CallState.swift` is now byte-identical to `develop`.

Going through `enqueueOwnCapabilities` also fixes the two CodeRabbit findings for free: the publisher is notified, revoked tracks are muted, and screensharing stops.

One thing I added that is worth your call: `enqueueOwnCapabilities` only took a no-arg closure, so applying an overlay meant reading `ownCapabilities` and writing it back across two actor hops. I added an overload that receives the current set, mirroring `enqueueCallSettings(_ operation: (CallSettings) -> CallSettings)`, so the read and write happen without suspension. The ~50 existing call sites are untouched. If you would rather I read-then-write the way `updateOwnCapabilities` already does, I will drop the overload.

Sources/StreamVideo/CallState.swift
```diff
@@ -501,13 +504,33 @@ public class CallState: ObservableObject {
...
+        ownCapabilities = callGrants.applied(to: ownCapabilities)
+    }
+
+    /// Stores the capabilities reported by the coordinator, with the latest SFU
+    /// grants applied on top.
+    ///
+    /// - Parameter value: The capabilities reported by the coordinator.
+    private func setOwnCapabilities(_ value: [OwnCapability]) {
+        ownCapabilities = callGrants?.applied(to: value) ?? value
+    }
```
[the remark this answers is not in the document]
martinmitrevski: Valid, and resolved in eb394e3e by removing the storage rather than adding clearing logic.

`CallState` no longer holds grants at all, so there is nothing to go stale and no `compactMap` swallowing the `nil` from `cleanUp()`. Grants are applied at the SFU boundary into the adapter's `ownCapabilities`, and `CallController` forwards that set to `Call.state`. `CallState.swift` is now byte-identical to `develop`.

Sources/StreamVideo/WebRTC/v2/WebRTCStateAdapter.swift
```diff
@@ -293,6 +296,9 @@ actor WebRTCStateAdapter: ObservableObject, StreamAudioSessionAdapterDelegate, W
     /// Sets the anonymous participant count.
     func set(anonymousCount value: UInt32) { self.anonymousCount = value }
 
+    /// Sets the publishing rights granted by the SFU.
+    func set(callGrants value: CallGrants?) { self.callGrants = value }
```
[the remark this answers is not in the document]
martinmitrevski: Correct, and resolved in eb394e3e: grants now go through `enqueueOwnCapabilities(_:)`, exactly as suggested.

```swift
let grants = event.currentGrants
await stateAdapter.enqueueOwnCapabilities { grants.applied(to: $0) }
```

So a revocation now notifies `publisher?.didUpdateOwnCapabilities`, turns off enabled audio/video, and stops active screensharing.

Worth noting for accuracy: on the previous revision the mute did eventually happen, but only indirectly. `Call+JoinedStage` observes `state.$ownCapabilities` and pushes back down through `updateOwnCapabilities`, so it needed a full round trip through the call state and only worked while the coordinator was in the `.joined` stage. It is direct now.

`WebRTCStateAdapter_Tests.test_enqueueOwnCapabilities_derivedFromCurrentSet_revokesSendAudioCapability_turnsAudioOff` covers the mute through the new overload.

martinmitrevski:
Pushed eb394e3e addressing all review comments. Replied in each thread; summary here.

**@ipavlidakis — grants should not be a passthrough property.** Adopted in full. `CallGrants` and `WebRTCStateAdapter.callGrants` are gone; `SFUEventAdapter` converts and enqueues, `CallController` observes `stateAdapter.$ownCapabilities` and forwards to `Call.state`. The net diff against `develop` is now smaller and `CallState.swift` is untouched.

**CodeRabbit, stale grants after cleanup.** Resolved by removing the storage instead of adding clearing logic. Nothing holds grants across a cleanup any more.

**CodeRabbit, revoked grant left publishing active.** Resolved by routing through `enqueueOwnCapabilities(_:)`, which notifies the publisher, mutes revoked tracks and stops screensharing.

**CodeRabbit, test naming.** Fixed the two `SFUEventAdapter_Tests` methods that carried the literal `given`/`when`/`then` words. Note the file's other ~15 tests use that style, so the new ones now read differently from their neighbours; AGENTS.md is explicit, so I followed it over local consistency. The other files flagged already used condition-action-result names.

**CodeRabbit, docstring coverage.** All three new declarations carry `///` docs.

### One trade-off to decide

This design applies grants into the capability set rather than keeping them as a persistent overlay, so **grants no longer survive a later coordinator capability update**. If `call.get()`, `call.update()` or an `UpdatedCallPermissionsEvent` lands after a revoke, its capabilities win.

That is a real behavioural difference from the JS SDK, which keeps grants in a separate subject and re-applies them on every capability change (`CallState.ts`, the `combineLatest` over `ownCapabilitiesSubject` and `callGrantsSubject`). Preserving that would mean storing the grants somewhere, which is the property this revision removes, so I have implemented it as requested rather than quietly keeping the overlay. Last-writer-wins is defensible, since SFU grants and coordinator permissions are two views of the same server-side truth and the SFU event normally follows the coordinator change. Happy to reintroduce the overlay if you would rather match JS exactly.

### Verification

- `WebRTCStateAdapter_Tests`, `SFUEventAdapter_Tests`, `Stream_Video_Sfu_Models_CallGrants_OwnCapability_Tests`, `CallController_Tests`, `CallState_Tests`, `Call_Tests`: all 8 new tests pass.
- Two unrelated tests (`CallState_Tests.test_didUpdate_performanceWithManyParticipants...` and `CallController_Tests.test_joinCall_whenPreviousAttemptFails...`) timed out on one local run while several test processes were competing for the machine; both pass in isolation, and neither touches code this PR changes.
- I also dropped an end-to-end mute test I had added at the SFU layer: `enqueueCallSettings` gates on capabilities and the settings queue made it slow and flaky in that fixture. The same behaviour is asserted at the adapter layer instead, where the existing infrastructure already covers it.

Sources/StreamVideo/WebRTC/v2/WebRTCStateAdapter.swift
```diff
@@ -790,6 +790,29 @@ actor WebRTCStateAdapter: ObservableObject, StreamAudioSessionAdapterDelegate, W
...
+    /// - Parameter operation: Receives the current capabilities and returns the
+    ///   new ones.
+    func enqueueOwnCapabilities(
+        functionName: StaticString = #function,
+        fileName: StaticString = #fileID,
+        lineNumber: UInt = #line,
+        _ operation: @Sendable @escaping (Set<OwnCapability>) -> Set<OwnCapability>
+    ) async {
+        let newValue = operation(ownCapabilities)
+        await enqueueOwnCapabilities(
```
ipavlidakis: How is that not causing a loop? enqueueOwnCapabilities -> enqueueOwnCapabilities -> enqueueOwnCapabilities -> ...

ipavlidakis: Because this one has a different closure signature than the original function. I'm not sure this is needed. Rather than having to model things so that we create ownCapabilities from grants which requres state read (to get the current capabilities) we can instead map callGrants into expected capabilities and pass  those into the adapter and let it perform the interesection for the 2 sets without requiring to expose a new state read.

martinmitrevski: It wasn'\''t recursing — the two are separate overloads and `{ newValue }` is a zero-parameter closure literal, which can only bind to the `() -> Set<OwnCapability>` overload (Swift refuses to implicitly discard a closure parameter, so the `(Set<OwnCapability>) -> Set<OwnCapability>` one is never a candidate).

That said, relying on overload resolution to prove there'\''s no loop is too subtle. Pushed f5befacd: both overloads now just evaluate their closure and call a shared private `apply(ownCapabilities:...)` that holds the side effects, so neither calls the other. Build and `WebRTCStateAdapter_Tests` pass.

ipavlidakis (approved):

PR #1251, opened 2026-08-19
Keep ParticipantSource across participant updates

martinmitrevski:
### 🔗 Issue Links

N/A — found while implementing participant reactions, where a neighbouring field would have hit the same trap.

### 🎯 Goal

`CallParticipant.withUpdated(...)` rebuilds the participant by listing every property, and `source` was never listed. Because the initializer defaults it to `.webRTCUnspecified`, every update silently dropped it.

So an RTMP, WHIP, SRT or RTSP participant lost its source the first time anything about them changed — a track being enabled, a track size update, becoming the dominant speaker. All 13 builders were affected.

The consequence is not cosmetic: `videoIngressSource` and `participantSource` are the comparators that put ingest participants first, and the **speaker**, **paginated** and **livestream/audio-room** sort presets all use `videoIngressSource`. Once a participant's source was reset, those comparators stopped recognising them, so ingest participants stopped being prioritised shortly after joining.

### 📝 Summary

- `source` is now carried forward by all 13 `withUpdated(...)` builders.
- `source` is now part of `CallParticipant` equality; previously two participants differing only by source compared equal.
- Two tests covering both.

### 🛠 Implementation

The fix follows the file's existing style: each builder lists the properties explicitly, so `source: source` is added to each of the 13. I deliberately did not restructure the builders into `var copy = self; copy.x = y; return copy` in this PR, even though that would make the whole class of bug impossible — it would turn a 13-line fix into a rewrite of a public model. Worth doing separately if you agree it's an improvement.

The equality change is the same omission seen from the other side. `source` was excluded from `==`, which matters now that it survives updates: without it, a participant whose source is corrected compares equal to one that still has the stale value, so `@Published` diffing and `removeDuplicates()` would drop the correction.

### 🧪 Manual Testing Notes

In a call with an RTMP or WHIP ingest participant, using the speaker or paginated layout: on `develop` the ingest participant is prioritised on join and then drifts down the list as soon as their tracks update. With this change they keep their position.

Automated coverage: `CallParticipants_Tests.test_withUpdated_preservesSource` checks five different builders, and the 16 existing `Sorting_Tests` cases for `videoIngressSource` / `participantSource` / `withParticipantSources` pass.

### ☑️ Contributor Checklist

- [ ] Affected documentation updated (tutorial, CMS) — N/A, restores documented behaviour

## Summary by CodeRabbit

* **Bug Fixes**
  * Participant comparisons now include source information for more accurate equality checks.
  * Participant updates preserve the existing source, preventing incorrect comparison results after updates.

* **Tests**
  * Added coverage for source preservation and equality behavior.

* **Documentation**
  * Added the fix to the upcoming changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### 🐞 Fixed
 - Transient peer-connection disconnections no longer trigger an immediate full rejoin, allowing the existing ICE restart flow to recover the session. [#1231](https://github.com/GetStream/stream-video-swift/pull/1231)
-- `CallParticipant.withUpdated(...)` no longer resets `source` to `.webRTCUnspecified`, which previously broke the `videoIngressSource` and `participantSource` sort comparators after any participant update. `source` is now also part of `CallParticipant` equality.
+- `CallParticipant.withUpdated(...)` no longer resets `source` to `.webRTCUnspecified`, which previously broke the `videoIngressSource` and `participantSource` sort comparators after any participant update. `source` is now also part of `CallParticipant` equality. [#1251](https://github.com/GetStream/stream-video-swift/pull/1251)
 
 # [1.51.0](https://github.com/GetStream/stream-video-swift/releases/tag/1.51.0)
 _August 13, 2026_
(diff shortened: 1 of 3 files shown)


ipavlidakis (approved):
Makes sense! 🚢

PR #1253, opened 2026-08-21
[Fix] Update ringing call state on reconnect

ipavlidakis:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1981/bugupdate-ringing-call-state-on-reconnect

### 🎯 Goal

After a WebSocket reconnect while an outgoing call is still ringing, accept/reject/end coordinator events can be missed. Local Call.state stays stale, so the caller never joins even though the callee already acted.

### 📝 Summary

- Reload state.ringingCall with get() (without ring: true) on every WSConnected.
- Replay accept/reject/end from the refreshed session into the existing CallViewModel handlers.
- Attach the session observer as soon as ringingCall is assigned, then check createdBy on session updates so CallKit-style get() hydration still works.
- Add adapter unit tests, including late createdBy hydration.

### 🛠 Implementation

- RingingCallRecoveryAdapter listens to rawEventPublisher for internal WSConnected events and refreshes the current ringing call from the coordinator.
- RingingFlowRecoveryAdapter watches StreamVideo.state.$ringingCall and synthesizes the same accept/reject/end events CallViewModel already handles from the socket.
- createdBy is checked when the session updates, not at ringing-call assignment, because CallKit assigns ringingCall and only then fills createdBy via get().

### 🎨 Showcase

Not applicable. No UI layout changes.

|  Before  |  After  |
| -------- | ------- |
|  N/A     |  N/A    |

### 🧪 Manual Testing Notes

1. Start an outgoing ringing call.
2. Drop the caller WebSocket while the callee is still ringing.
3. Have the callee accept, reject, or let the call end.
4. Restore the caller connection.
5. Confirm the caller leaves the ringing UI and joins, hangs up, or ends according to the refreshed session instead of staying stuck.

### ☑️ Contributor Checklist

### 🎁 Meme

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ringing call recovery after WebSocket reconnections.
  * Accept, reject, and end actions now reflect refreshed call state, preventing stale ringing calls.
  * Corrected SFU telemetry durations after fast reconnects.
  * Fixed post-call feedback session IDs.
  * Updated participant sorting to prioritize screensharing over pinned participants.

* **Tests**
  * Added coverage for reconnect recovery across outgoing and incoming ringing call scenarios.


Sources/StreamVideo/Utils/RingingCallRecoveryAdapter/RingingCallRecoveryAdapter.swift
```diff
@@ -0,0 +1,63 @@
...
+        guard
+            let ringingCall = await MainActor.run(body: {
+                streamVideo?.state.ringingCall
+            })
+        else {
+            return
+        }
+
+        do {
+            _ = try await ringingCall.get()
```
martinmitrevski, line 58: is this enough to update the state? Shouldn't we explicitly do that?

martinmitrevski: ok, checked this, nothing else is needed.

martinmitrevski (approved):

PR #1254, opened 2026-08-24
Report user session id with collected feedback

martinmitrevski:
Addresses the D7 finding from the cross-SDK consistency audit on iOS.

`CollectUserFeedbackRequest` has a `userSessionId` field that the SDK never set, so feedback submitted from iOS could not be joined to the call session it refers to — or to that session's stats. Both other platforms report it: JS sends `user_session_id`, Android passes `sessionId`.

### Why not read the session id at submission time

`WebRTCStateAdapter.cleanUp()` resets the session id to `""` while the call is torn down, reached on every call end via `leave()` → Leaving → CleanUp, and directly via `CallController.cleanUp()`. User feedback is typically collected *after* that, from a post-call rating screen — the pattern in our own `18-call-quality-rating.swift` docs. A live read therefore ships `"user_session_id": ""` in exactly the case the finding is about.

So `CallController` retains the latest non-empty session id from the observer that already tracks session-id changes, and reports that.

### Notes

- The id is available as long as feedback is submitted from the same `Call` instance that joined. An app that re-creates the call via `streamVideo.call(callType:callId:)` for its rating screen gets a fresh controller and no id. JS is instance-scoped in the same way.
- `@Atomic` is needed because `lastSessionId` is written on `@MainActor` and read from the `nonisolated async` `collectUserFeedback`.

### Left out of scope

Two parts of D7 that are cross-SDK decisions rather than iOS-side fixes:

- The `sdk` identifier naming split — `stream-ios` vs `stream-js`/`stream-react` vs `stream-video-android`. `SystemEnvironment.sdkName` rides on every client event, not just feedback, so renaming it unilaterally would break existing dashboards.
- JS's `x-stream-platform-data` entry inside `custom`. It's 1-of-3, and injecting SDK data into the user-supplied `custom` dictionary risks key collisions in a payload apps control.

### Testing

Two tests added to `CallController_Tests`, both driving real paths — `test_collectUserFeedback_afterCallEnded_reportsSessionIdOfEndedSession` exercises the actual `cleanUp()` teardown and fails against a live read (`("Optional("")") is not equal to ("Optional("2F59B930-…")")`). Full class: 47 passed, 0 failed.

## Summary by CodeRabbit

* **Bug Fixes**
  * User feedback now includes the call’s latest WebRTC session ID for accurate backend correlation.
  * Session information remains available when feedback is submitted after call cleanup.
  * Empty session updates no longer overwrite the retained session ID.

* **Documentation**
  * Updated the upcoming changelog with the feedback session ID improvement.

### DIFF
--- a/Sources/StreamVideo/Controllers/CallController.swift
+++ b/Sources/StreamVideo/Controllers/CallController.swift
@@ -76,6 +76,14 @@ class CallController: @unchecked Sendable {
     private var webRTCParticipantsObserver: AnyCancellable?
     private var participants: CollectionDelayedUpdateObserver<[String: CallParticipant]>?
 
+    /// The latest non-empty session id observed for this call.
+    ///
+    /// `WebRTCStateAdapter.cleanUp()` resets the session id while the call is
+    /// torn down, but user feedback is usually collected *after* the call has
+    /// ended. Retaining the id here keeps it available to
+    /// ``collectUserFeedback(custom:rating:reason:)``.
+    @Atomic private var lastSessionId: String?
+
     private let disposableBag = DisposableBag()
 
     init(
@@ -479,6 +487,11 @@ class CallController: @unchecked Sendable {
 
     /// Collects user feedback asynchronously.
     ///
+    /// The submission carries the call's session id, which allows the feedback
+    /// to be correlated with the call session (and its stats) on the backend.
+    /// It is reported even when the feedback is collected after the call
+    /// ended, which is the common case for a post-call rating screen.
+    ///
     /// - Parameters:
     ///   - custom: Optional custom data in the form of a dictionary of String keys and RawJSON values.
     ///   - rating: Optional rating provided by the user.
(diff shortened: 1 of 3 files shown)


oliverlaz (approved):

Sources/StreamVideo/Controllers/CallController.swift
```diff
@@ -76,6 +76,14 @@ class CallController: @unchecked Sendable {
     private var webRTCParticipantsObserver: AnyCancellable?
     private var participants: CollectionDelayedUpdateObserver<[String: CallParticipant]>?
 
+    /// The latest non-empty session id observed for this call.
+    ///
+    /// `WebRTCStateAdapter.cleanUp()` resets the session id while the call is
+    /// torn down, but user feedback is usually collected *after* the call has
+    /// ended. Retaining the id here keeps it available to
+    /// ``collectUserFeedback(custom:rating:reason:)``.
+    @Atomic private var lastSessionId: String?
```
ipavlidakis, line 85: I would prefer to move that into the StateAdapter rather than having the CallController shadowing a new property related to session. While the StateAdapter is expected to remain stateful and provide that kind of information.

PR #1255, opened 2026-08-24
Report fast reconnect duration in join telemetry

martinmitrevski:
Addresses the D5 finding from the cross-SDK consistency audit — flagged High, and the audit's "fix this first" item.

`JoinedStateTelemetryReporter` populates a `Reconnection` value with the elapsed duration, but the `.fast` branch declared a *second* local `reconnection` that shadowed it and set only `strategy`. The shadowed value is what got sent, so `timeSeconds` was `0` for every fast reconnect from iOS — structurally zero, silently dragging any cross-platform duration average down. `.rejoin` and `.migrate` mutate the outer value and were always correct, which is why this survived review.

JS and Android both send strategy and time together in one payload, so this is a Swift-only defect rather than a design difference.

```diff
             case .fast:
-                var reconnection = Stream_Video_Sfu_Signal_Reconnection()
                 reconnection.strategy = .fast
                 return .reconnection(reconnection)
```

### Testing

`JoinedStateTelemetryReporter` had no test coverage at all. Added `JoinedStateTelemetryReporter_Tests` covering all four flow types plus session-id forwarding, and verified the `.fast` case genuinely guards the bug — with the fix reverted it fails as:

```
JoinedStateTelemetryReporter_Tests.swift:35: XCTAssertGreaterThan failed: ("0.0") is not greater than ("0.0")
```

while `regular`, `rejoin` and `migrate` stay green, matching the finding's scope.

The three existing `reportsTelemetry` tests in `WebRTCCoordinatorStateMachine_JoiningStageTests` asserted `reconnection.strategy` but never `timeSeconds` — the gap that let a zero duration through the real state-machine flow. They now assert both, and all three pass.

### Note on an unrelated failing test

`test_willTransitionAway_audioReadinessPending_skipsPeerConnections` fails in a full-class run of `WebRTCCoordinatorStateMachine_JoiningStageTests`. It is **pre-existing and unrelated**: it fails the same way on clean `develop` with these changes stashed (52 passed / 1 failed), and passes when run in isolation, so it is ordering-dependent.

## Summary by CodeRabbit

* **Bug Fixes**
  * SFU telemetry now reports the actual elapsed duration for fast reconnect, rejoin, and migration events instead of zero.
  * Connection telemetry now consistently includes reconnection details and session identifiers.

* **Tests**
  * Added coverage verifying connection timing, reconnection strategies, session identifiers, and positive duration values.

### DIFF
--- a/Sources/StreamVideo/WebRTC/v2/StateMachine/Components/JoinedStateTelemetryReporter.swift
+++ b/Sources/StreamVideo/WebRTC/v2/StateMachine/Components/JoinedStateTelemetryReporter.swift
@@ -40,7 +40,6 @@ struct JoinedStateTelemetryReporter {
             case .regular:
                 return .connectionTimeSeconds(duration)
             case .fast:
-                var reconnection = Stream_Video_Sfu_Signal_Reconnection()
                 reconnection.strategy = .fast
                 return .reconnection(reconnection)
             case .rejoin:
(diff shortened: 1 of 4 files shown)


oliverlaz (approved):

Sources/StreamVideo/WebRTC/v2/StateMachine/Components/JoinedStateTelemetryReporter.swift
```diff
@@ -40,7 +40,6 @@ struct JoinedStateTelemetryReporter {
             case .regular:
                 return .connectionTimeSeconds(duration)
             case .fast:
-                var reconnection = Stream_Video_Sfu_Signal_Reconnection()
```
ipavlidakis, line 43: Great catch!

PR #1258, opened 2026-08-28
[v2] Add the shared design token system to the Video SDK

nuno-vieira:
Puts the Video SDK on `DesignSystemTokens` from `StreamCoreUI`, so the v2 UI redesign is built on the same primitives as the rest of Stream's SDKs. This PR introduces `VideoAppearance` and the shared color and layout tokens. Existing views stay on `Appearance` until they migrate.

> [!NOTE]
> Draft until [GetStream/stream-core-swift#81](https://github.com/GetStream/stream-core-swift/pull/81) merges and ships. `Package.swift` is pinned to the `core-ui-design-tokens` branch and needs repointing to a released version before this can land.

## Architecture

`Appearance` is unchanged for views: `colors`, `images`, `fonts`, `sounds`, and `StreamVideoUI`. `VideoAppearance` is the new design-system type. Shared color and layout tokens come from a `DesignSystemTokens` instance passed into it. Video-only colors are the nested type `VideoAppearance.Colors` and are read on `videoAppearance.colors`.

```mermaid
flowchart TB
  subgraph core ["StreamCoreUI"]
    DT["DesignSystemTokens"]
    DT --> C["Colors"]
    DT --> L["Layout"]
  end

  subgraph video ["StreamVideoSwiftUI"]
    A["Appearance\ncolors, images, fonts, sounds"]
    VA["VideoAppearance"]
    VA --> DT
    VA --> VC["colors\nVideoAppearance.Colors"]
    VC --> C
  end
```

```swift
let tokens = DesignSystemTokens()
tokens.colors.accentPrimary = .red
let videoAppearance = VideoAppearance(tokens: tokens)
videoAppearance.colors.indicatorSpeaking = .green
```

Pass the same `DesignSystemTokens` instance into Chat later so both SDKs reskin together. Color tokens derive lazily, so override ramps before the first read.

## Testing

`Appearance_DesignSystem_Tests` covers isolated instances, shared `DesignSystemTokens` identity, and the cascade from a shared ramp into a video token.

## Follow-ups

- Chat adoption: [IOS-1999](https://linear.app/stream/issue/IOS-1999)
- Moving the token scope split into the generator: [IOS-2000](https://linear.app/stream/issue/IOS-2000)

### DIFF
--- a/.github/workflows/smoke-checks.yml
+++ b/.github/workflows/smoke-checks.yml
@@ -167,6 +167,8 @@ jobs:
     if: ${{ github.event.inputs.record_snapshots_swiftui != 'true' && github.event.inputs.record_snapshots_uikit != 'true' }}
     steps:
     - uses: actions/checkout@v4.1.1
+      with:
+        fetch-depth: 0
     - uses: ./.github/actions/bootstrap
       env:
         INSTALL_INTERFACE_ANALYZER: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+# 2.0.0
+
+### ✅ Added
+- New `DesignSystemTokens` in `StreamCoreUI` is used in `VideoAppearance` for primitive UI tokens [#1258](https://github.com/GetStream/stream-video-swift/pull/1258)
+
+### 🔄 Changed
+- The `Appearance` object has been replaced with `VideoAppearance` object [#1258](https://github.com/GetStream/stream-video-swift/pull/1258)
+
 # Upcoming
 
 ### ✅ Added
--- a/Sources/StreamVideoSwiftUI/DesignSystem/VideoAppearance+Colors.swift
+++ b/Sources/StreamVideoSwiftUI/DesignSystem/VideoAppearance+Colors.swift
@@ -1,0 +1,48 @@
+//
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+import StreamCoreUI
+import UIKit
+
+/// Video-specific color tokens, derived from ``DesignSystemTokens/Colors``.
+///
+/// Read them on the Video appearance:
+/// `videoAppearance.colors.indicatorSpeaking`.
+extension VideoAppearance {
+    public final class Colors {
+        private let colors: DesignSystemTokens.Colors
+
+        // MARK: - Control
+
+        public lazy var controlAcceptCallBackground: UIColor = colors
+            .accentSuccess
+        public lazy var controlAcceptCallText: UIColor = colors.textOnAccent
+        public lazy var controlVideoBackgroundControlBackground: UIColor = colors
+            .backgroundCoreSurfaceSubtle
+        public lazy var controlVideoBackgroundControlBackgroundSelected: UIColor =
+            colors.accentPrimary
+        public lazy var controlVideoBackgroundControlText: UIColor = colors
+            .textPrimary
+        public lazy var controlVideoBackgroundControlTextSelected: UIColor =
+            colors.textOnAccent
+
+        // MARK: - Indicator
+
+        public lazy var indicatorFair: UIColor = colors.accentWarning
+        public lazy var indicatorGreat: UIColor = colors.accentSuccess
+        public lazy var indicatorPoor: UIColor = colors.accentError
+        public lazy var indicatorSpeaking: UIColor = colors.palette.brand300
+
+        // MARK: - Label
+
+        public lazy var labelBackgroundNeutral: UIColor = colors.palette.chrome150
+        public lazy var labelBackgroundPrimary: UIColor = colors.palette.brand150
+        public lazy var labelTextNeutral: UIColor = colors.textPrimary
+        public lazy var labelTextPrimary: UIColor = colors.palette.brand900
+
+        public init(tokens: DesignSystemTokens = DesignSystemTokens()) {
+            self.colors = tokens.colors
+        }
+    }
+}
--- a/Sources/StreamVideoSwiftUI/Fonts.swift
+++ b/Sources/StreamVideoSwiftUI/Fonts.swift
@@ -5,7 +5,7 @@
 import SwiftUI
 
 /// Provides access to fonts used in the SDK.
-public struct Fonts {
+public struct Fonts: Sendable {
     
     public init() { /* Public init. */ }
     
--- a/Sources/StreamVideoSwiftUI/VideoAppearance.swift
+++ b/Sources/StreamVideoSwiftUI/VideoAppearance.swift
@@ -1,0 +1,39 @@
+//
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+@_exported import StreamCoreUI
+import StreamVideo
+import SwiftUI
+
+/// Video design-system configuration.
+///
+/// Shared color and layout tokens come from ``DesignSystemTokens``. Pass
+/// the same instance into Chat's appearance so both SDKs reskin together.
+/// Video-only colors live on ``colors``. Existing views still use
+/// ``Appearance`` until they migrate.
+///
+/// ```swift
+/// let tokens = DesignSystemTokens()
+/// tokens.colors.accentPrimary = .red
+/// let appearance = VideoAppearance(tokens: tokens)
+/// appearance.colors.indicatorSpeaking = .green
+/// ```
+public final class VideoAppearance {
+    /// The instance the Video SDK uses unless it is given another one.
+    ///
+    /// Not synchronized. This is a process-wide UI configuration object.
+    public nonisolated(unsafe) static let shared = VideoAppearance()
+
+    /// Shared color and layout tokens. Mutating this instance is visible
+    /// to any other appearance constructed with it.
+    public let tokens: DesignSystemTokens
+
+    /// Video-specific colors, derived from ``tokens``.
+    public var colors: Colors
+
+    public init(tokens: DesignSystemTokens = DesignSystemTokens()) {
+        self.tokens = tokens
+        self.colors = Colors(tokens: tokens)
+    }
+}
--- a/.github/actions/xcode-cache/action.yml
+++ b/.github/actions/xcode-cache/action.yml
@@ -10,5 +10,4 @@ runs:
       id: spm-cache
       with:
         path: spm_cache
-        key: ${{ env.IMAGE }}-spm-${{ hashFiles('**/Package.resolved') }}
-        restore-keys: ${{ env.IMAGE }}-spm-
+        key: ${{ env.IMAGE }}-spm-${{ hashFiles('Package.swift') }}
--- a/DocumentationTests/DocumentationTests/DocumentationTests/04-ui-components/03-video-theme.swift
+++ b/DocumentationTests/DocumentationTests/DocumentationTests/04-ui-components/03-video-theme.swift
@@ -10,6 +10,13 @@ import SwiftUI
 @MainActor
 private func content() {
     container {
+        let tokens = DesignSystemTokens()
+        tokens.colors.accentPrimary = .red
+        let appearance = VideoAppearance(tokens: tokens)
+        appearance.colors.indicatorSpeaking = .green
+    }
+
+    container {
         let streamBlue = UIColor(red: 0, green: 108.0 / 255.0, blue: 255.0 / 255.0, alpha: 1)
         var colors = Colors()
         colors.tintColor = Color(streamBlue)
--- a/Package.swift
+++ b/Package.swift
@@ -24,7 +24,12 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-protobuf.git", exact: "1.30.0"),
         .package(url: "https://github.com/GetStream/stream-video-swift-webrtc.git", exact: "145.15.0"),
-        .package(url: "https://github.com/GetStream/stream-core-swift.git", from: "0.10.1")
+        // Temporary pin to the develop commit that added StreamCoreUI
+        // tokens. Restore `from:` once they ship in a tag.
+        .package(
+            url: "https://github.com/GetStream/stream-core-swift.git",
+            revision: "7dcc7c889dcd1d860576c82cac8a0f486cec7d26"
+        )
     ],
     targets: [
         .target(
@@ -37,7 +42,10 @@ let package = Package(
         ),
         .target(
             name: "StreamVideoSwiftUI",
-            dependencies: ["StreamVideo"],
+            dependencies: [
+                "StreamVideo",
+                .product(name: "StreamCoreUI", package: "stream-core-swift")
+            ],
             resources: [
                 .process("Resources")
             ]
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ You can customize the look and feel of the screens presented in the calling flow
 
 Most of our components are public, so you can use them as building blocks if you want to build your custom UI.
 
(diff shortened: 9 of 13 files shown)


Sources/StreamVideoSwiftUI/DesignSystem/Appearance+ColorPalette.swift
```diff
@@ -0,0 +1,435 @@
...
+    ///
+    /// - Important: Start by adjusting the `brand` and `chrome` ramps. Most
+    /// semantic tokens derive from them, so a handful of overrides reskins the
+    /// whole SDK. Reach for an individual token only when one surface needs to
+    /// deviate from what the ramps produce.
+    ///
+    /// Because tokens derive lazily, override the ramps before the first read.
+    /// Assigning to `Appearance.colorPalette` on `StreamVideoUI` setup is the
+    /// safest point to do so.
+    final class ColorPalette: @unchecked Sendable {
```
martinmitrevski: shouldn't this be pulled from the common module?

martinmitrevski (commented):
I don't see much point in this PR tbf. We need to move the common product agnostic tokens from chat common UI to stream core UI, and pull them here and start replacing them in the views.

Sources/StreamVideoSwiftUI/DesignSystem/Appearance+VideoColors.swift
```diff
@@ -0,0 +1,55 @@
...
+    // MARK: - Indicator
+
+    public lazy var indicatorFair: UIColor = palette.accentWarning
+    public lazy var indicatorGreat: UIColor = palette.accentSuccess
+    public lazy var indicatorPoor: UIColor = palette.accentError
+    public lazy var indicatorSpeaking: UIColor = palette.brand300
+
+    // MARK: - Label
+
+    public lazy var labelBackgroundNeutral: UIColor = palette.chrome150
```
ipavlidakis: Why UIColor? Wouldn't it make more sense to use the cross-platform SwiftUI.Color instead so we are not blocked on supporting things like AppKit?

Sources/StreamVideoSwiftUI/Fonts.swift
```diff
@@ -2,41 +2,31 @@
...
+///
+/// - Note: This is a reference type, so a copy is not a snapshot: mutating
+/// one reference is visible through every other reference to the same
+/// instance, and through the SDKs that share it.
+@dynamicMemberLookup
+public final class Fonts: @unchecked Sendable {
+    /// The styles this SDK shares with the rest of Stream's SDKs.
+    ///
+    /// Pass a separate instance to give this SDK its own.
+    public let shared: SharedFonts.SwiftUIFace
```
ipavlidakis: Does that need to be public (and the init below)? Integrators can always update values via the dynamicScript. If there is no reason to expose directly `SharedFonts.SwiftUIFace` i would prefer to optin for making the existing initialiser, internal and create a public convenience without any parameters that will pass the sharedValue while also making the shared value private.

nit: I would also name it `source` rather than shared

Sources/StreamVideoSwiftUI/DesignSystem/Appearance+VideoColors.swift
nuno-vieira: That is a good point. For video SDK colors only it could make sense to use SwiftUI.Color, but it would not be consistent with the rest of the colors. They are all UIColor to support the Chat UIKit SDK. But yes this means that AppKit would not be supported

Sources/StreamVideoSwiftUI/Fonts.swift
nuno-vieira: TBH, AI kinda went rogue here and also did Fonts. I'm planning on setting this up in a seperate PR, so I didn't pay attention to how it is done at the moment, I will remove it from here, and then we can discuss it better. But Fonts on Chat were tricky because we had to have 2 seperate objects, one for SwiftUI and one for UIKit. In this case, we won't need the UIKit one, so things can be simpler

Sources/StreamVideoSwiftUI/VideoAppearance.swift
```diff
@@ -0,0 +1,42 @@
...
+/// appearance.colors.indicatorSpeaking = .green
+/// ```
+public final class VideoAppearance {
+    /// The instance the Video SDK uses unless it is given another one.
+    ///
+    /// Not synchronized. This is a process-wide UI configuration object.
+    public nonisolated(unsafe) static let shared = VideoAppearance()
+
+    /// Same instance as ``shared``.
+    public static var `default`: VideoAppearance { shared }
```
ipavlidakis: why do we need both shared and default?

Sources/StreamVideoSwiftUI/VideoAppearance.swift
```diff
@@ -0,0 +1,42 @@
...
+
+    /// Shared colour and layout tokens. Mutating this instance is visible
+    /// to any other appearance constructed with it.
+    public let tokens: DesignSystemTokens
+
+    /// Video-specific colours, derived from ``tokens``.
+    public var colors: Colors
```
ipavlidakis, line 33: As i mentioned in the [comment](https://github.com/GetStream/stream-core-swift/pull/81#discussion_r3902454829) in the Core PR i think this one should also be changed.

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+# 2.0.0
```
ipavlidakis, line 5: We aren't probably targeting for 2.0, right 😅 ?

Sources/StreamVideoSwiftUI/DesignSystem/VideoAppearance+Colors.swift
```diff
@@ -0,0 +1,48 @@
...
+import StreamCoreUI
+import UIKit
+
+/// Video-specific colour tokens, derived from ``DesignSystemTokens/Colors``.
+///
+/// Read them on the Video appearance:
+/// `videoAppearance.colors.indicatorSpeaking`.
+extension VideoAppearance {
+    public final class Colors {
+        private let colors: DesignSystemTokens.Colors
```
martinmitrevski, line 14: maybe we call them baseColors or something like that?

Sources/StreamVideoSwiftUI/DesignSystem/VideoAppearance+Colors.swift
```diff
@@ -0,0 +1,48 @@
...
+        public lazy var indicatorSpeaking: UIColor = colors.brand300
+
+        // MARK: - Label
+
+        public lazy var labelBackgroundNeutral: UIColor = colors.chrome150
+        public lazy var labelBackgroundPrimary: UIColor = colors.brand150
+        public lazy var labelTextNeutral: UIColor = colors.textPrimary
+        public lazy var labelTextPrimary: UIColor = colors.brand900
+
+        public init(tokens: DesignSystemTokens) {
```
martinmitrevski: we should have a default value here

Sources/StreamVideoSwiftUI/VideoAppearance.swift
martinmitrevski: yes, I want to know that as well

CHANGELOG.md
martinmitrevski: We are - that's the plan for autumn

Sources/StreamVideoSwiftUI/VideoAppearance.swift
nuno-vieira: I don't know either; I assumed that video was using both? Probably AI got confused because on Chat we use default, and here we use shared. So I guess we should decide. Go for shared or default. The initial goal with default is that customers could create multiple appearances and access them like so:

```swift
Appearance.default // Default theme
Appearance.dark // Dark theme
Appearance.light // Light theme
```

But `default` is a bit of a misleading name, since a customer can customise it, so it is not really the default. I would probably go with shared, since it is clear that it is a singleton. 

This does not work at all like Apple's `UINavigationBar.appearance()` etc... I guess the first team was inspired by this, and decided to use the same name, but they work different.

Sources/StreamVideoSwiftUI/VideoAppearance.swift
nuno-vieira: Replied in Core

Sources/StreamVideoSwiftUI/DesignSystem/VideoAppearance+Colors.swift
nuno-vieira: They are private, so we can change that later, but they are accessed by `tokens.colors`, so it is clear that they are the foundation already

Sources/StreamVideoSwiftUI/DesignSystem/VideoAppearance+Colors.swift
nuno-vieira: Yes makes sense

nuno-vieira: Although, in one way, it could hide that these tokens exist. And customers need to know about it so that they can update the common stuff

Sources/StreamVideoSwiftUI/VideoAppearance.swift
```diff
@@ -0,0 +1,42 @@
...
+    ///
+    /// Not synchronized. This is a process-wide UI configuration object.
+    public nonisolated(unsafe) static let shared = VideoAppearance()
+
+    /// Same instance as ``shared``.
+    public static var `default`: VideoAppearance { shared }
+
+    /// Shared color and layout tokens. Mutating this instance is visible
+    /// to any other appearance constructed with it.
+    public let tokens: DesignSystemTokens
```
laevandus, line 30: So I access colors like this now, right?

` = videoAppearance.default.tokens.colors.accentPrimary`
and
` = videoAppearance.default.colors.controlAccept...`

Thinking if StreamCore Colors would be subclassable, we could access them the same way. 🤔

nuno-vieira: Subclassing them makes everything more difficult to find and access; it is hard to tell which tokens belong to what.

Customers will access this through the InjectedValues wrapper, so it will actually be only this:

```swift
@Injected(\.tokens) var tokens
@Injected(\.colors) var colors

tokens.colors.accentPrimary
colors.controlAccept
```

Another options is move the tokens colors inside the video appearace colors:

```swift
@Injected(\.colors) var colors

colors.tokens.accentPrimary
colors.controlAccept
```

Sources/StreamVideoSwiftUI/DesignSystem/VideoAppearance+Colors.swift
nuno-vieira: This is already landed: `VideoAppearance.Colors.init` defaults to `DesignSystemTokens()`, so `Colors()` works without supplying tokens.

Sources/StreamVideoSwiftUI/VideoAppearance.swift
nuno-vieira: Still not 100% happy with this. Becuase if a customer uses both Chat and Video SDKs, then how do you access the "common" colors? Do you go always through VideoAppearance, or ChatAppearance? Or would customers directly go through the Tokens object?

nuno-vieira: The best approach is probably to keep it like it is now, the InjectedValues will then be the appearance object it self. Since if a customer is using both SDKs, there is a need to have separate appearance objects. But when creating a custom component, that component is most likely on referencing one SDK, although you could opt to read both appearances at the same time. Example:

```swift
struct AcceptCallButton: View {
    @Injected(\.videoAppearance) private var appearance

    var body: some View {
        Circle()
            .fill(Color(appearance.colors.controlAcceptCallBackground))
            .padding(appearance.tokens.layout.spacingMd)
            .overlay {
                Circle()
                    .stroke(
                        Color(appearance.tokens.colors.accentPrimary),
                        lineWidth: appearance.tokens.layout.strokeSm
                    )
            }
    }
}

struct UnreadBadge: View {
    @Injected(\.chatAppearance) private var appearance

    var body: some View {
        Text("2")
            .foregroundColor(Color(appearance.tokens.colors.textPrimary))
            .padding(.horizontal, appearance.tokens.layout.spacingXs)
            .padding(.vertical, appearance.tokens.layout.spacing2xs)
            .background(Color(appearance.tokens.colors.accentPrimary))
            .clipShape(
                RoundedRectangle(
                    cornerRadius: appearance.tokens.layout.radiusMd
                )
            )
    }
}

struct InboxHeader: View {
    @Injected(\.videoAppearance) private var videoAppearance
    @Injected(\.chatAppearance) private var chatAppearance

    var body: some View {
        HStack(spacing: videoAppearance.tokens.layout.spacingMd) {
            Label("3 missed", systemImage: "phone.down.fill")
                .foregroundColor(
                    Color(videoAppearance.tokens.colors.textOnAccent)
                )
                .padding(videoAppearance.tokens.layout.spacingXs)
                .background(
                    Color(videoAppearance.colors.controlAcceptCallBackground)
                )
                .clipShape(
                    RoundedRectangle(
                        cornerRadius: videoAppearance.tokens.layout.radiusMd
                    )
                )

            Label("2 unread", systemImage: "message.fill")
                .foregroundColor(
                    Color(chatAppearance.tokens.colors.textPrimary)
                )
                .padding(chatAppearance.tokens.layout.spacingXs)
                .background(
                    Color(chatAppearance.tokens.colors.accentPrimary)
                )
                .clipShape(
                    RoundedRectangle(
                        cornerRadius: chatAppearance.tokens.layout.radiusMd
                    )
                )
        }
        .padding(videoAppearance.tokens.layout.spacingMd)
    }
}
```

Sources/StreamVideoSwiftUI/DesignSystem/VideoAppearance+Colors.swift
martinmitrevski: my thinking was just for us - we have a Colors class, and inside colors property. But anyway, not a blocker.

Sources/StreamVideoSwiftUI/VideoAppearance.swift
martinmitrevski: we can do a breaking change here as well

Sources/StreamVideoSwiftUI/VideoAppearance.swift
martinmitrevski: the suggested way imo should be via the tokens, and then inject the same object (if you need it), on both chat and video.

martinmitrevski (approved):
LGTM

.github/workflows/smoke-checks.yml
```diff
@@ -167,6 +167,8 @@ jobs:
     if: ${{ github.event.inputs.record_snapshots_swiftui != 'true' && github.event.inputs.record_snapshots_uikit != 'true' }}
     steps:
     - uses: actions/checkout@v4.1.1
+      with:
+        fetch-depth: 0
```
ipavlidakis: what is this about?

ipavlidakis (approved):

.github/workflows/smoke-checks.yml
nuno-vieira: I think it is because we have 30 commits on the PR right now, and it automated code review is failing because it can't fetch more than whatever number of commits

Sources/StreamVideoSwiftUI/VideoAppearance.swift
nuno-vieira: WDYM via the tokens?

PR #1260, opened 2026-09-01
[Enhancement]Music mode implementation

ipavlidakis:
### 🔗 Issue Links

In-call music / hi-fi capture.

### 🎯 Goal

Allow a joined call to switch capture from voice processing to a music / hi-fi path without leaving and rejoining:

```swift
try await call.microphone.setAudioBitrateProfile(.musicHighQuality)
```

Requires dashboard `hifi_audio_enabled`. Published as `call.microphone.audioBitrateProfile`. Persists across reconnect; resets on leave.

### 📝 Summary

- Public `AudioBitrateProfile`: `.voiceStandard` (64 kbps), `.voiceHighQuality` (128 kbps, processing on), `.musicHighQuality` (128 kbps, speech processing off).
- Music leaves VoiceChat (`playAndRecord` + `.default`) so Apple allows Voice Processing to be **disabled**, not only bypassed. Software NS/HPF off. Stream NC filter stashed and restored.
- Mute while music uses `.inputMixer`; never-music mute still matches `develop` (`.voiceProcessing`, `unpublish()`).
- 128 kbps is stamped on RTP params (no renegotiation) and re-stamped on reconnect.
- Applicator is opt-in: never-music join/leave does not touch session, ADM VP, APM, or bitrate.
- Demo debug more-controls: Music Mode toggle.

### 🔧 Implementation

**Reviewer file order** (do not start in `AudioDeviceModule.swift`):

1. `Sources/StreamVideo/CallSettings/AudioBitrateProfile.swift` — public enum.
2. `Sources/StreamVideo/CallSettings/MicrophoneManager.swift` — API + published value + leave reset.
3. `Sources/StreamVideo/WebRTC/v2/AudioBitrateProfileApplicator.swift` — apply owner (`setProfile` vs `rebind`, `.none` / `.bitrateOnly` / `.processing`).
4. `Sources/StreamVideo/Utils/AudioSession/AudioDeviceModule/MusicCapturePolicy.swift` — desired vs applied vs pending VP.
5. `Sources/StreamVideo/Utils/AudioSession/CallAudioSession.swift` — VoiceChat leave; `set` rolls back, `commit` does not.
6. `Sources/StreamVideo/WebRTC/v2/PeerConnection/MediaAdapters/LocalMediaAdapters/LocalAudioMediaAdapter.swift` — bitrate stored and stamped on new transceivers.
7. `Sources/StreamVideo/WebRTC/v2/PeerConnection/MediaAdapters/ScreenShareAudioFilterGate.swift` + capturer attach — screenshare does not write the shared filter itself.
8. Tests next to those types.

**Invariants**

- Never-music mute = `develop`: `setMicrophoneMuted` only; mute still `unpublish()`s.
- `.voiceHighQuality` is bitrate-only.
- Stereo preference only **bypasses** VP. Music-exit with stereo preferred records `.voiceWhileStereo` and restores VP when stereo turns off.
- Apple VP is one bundle (AEC+NS+AGC). Music turns it all off. Echo on speaker until software AEC — see the `ponytail:` comment on ADM.

### 🎨 Showcase

Debug more-controls: Music Mode toggle. Dashboard refusal is shown on the button label.

### 🧪 Manual Testing Notes

Dashboard Hi-fi on, two devices:

- [ ] Join → Music Mode → listener hears it → voice restores processing.
- [ ] Toggle twice without leaving.
- [ ] Mute → music → unmute (capture resumes).
- [ ] NC on while music; capture stays clean.
- [ ] Speaker → built-in receiver; session stays on `.default` while music is on.
- [ ] Leave + join on a cached `Call`; music can be set again.
- [ ] Never-music call: mute/unmute unchanged vs `develop`.

### ☑️ Contributor Checklist

### 🎁 Meme

_n/a_

## Summary by CodeRabbit

* **New Features**
  * Added selectable in-call audio profiles for standard voice, high-quality voice, and high-quality music.
  * Added a demo control for switching between voice and music audio modes.
  * Music mode provides higher-fidelity capture and increased bitrate when hi-fi audio is enabled in call settings.
* **Bug Fixes**
  * Audio filters and voice-processing settings now restore correctly after music mode or screen sharing.
  * Audio settings reset to standard voice mode when leaving a call.
  * Improved audio behavior during mute, unmute, reconnection, and audio-route changes.
  * Disabling noise cancellation now reliably clears active audio filters.


Sources/StreamVideo/CallStateMachine/Stages/Call+LeavingStage.swift
```diff
@@ -81,6 +81,9 @@ extension Call.StateMachine.Stage {
             input.resetAudioFilter()
 
             Task(disposableBag: disposableBag) { @MainActor [weak self, call] in
+                // Cached Call + leave + join: clear published music so
+                // the next set is not a no-op at the manager guard.
+                await call.microphone.resetAudioBitrateProfile()
```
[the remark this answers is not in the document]
ipavlidakis: Moved `resetAudioBitrateProfile()` to after the idle transition and `callEnded` so an in-flight music apply cannot delay teardown. It still runs on `audioBitrateProfileQueue`, so a cached Call rejoin is not a music no-op. Fixed in 4813ca4d.

Sources/StreamVideo/Controllers/CallController.swift
```diff
@@ -212,6 +212,20 @@ class CallController: @unchecked Sendable {
...
+    ///
+    /// Requires dashboard `hifi_audio_enabled`. Fetches call settings when
+    /// they have not been loaded yet. The coordinator/applicator owns
+    /// session, ADM, and stored profile; a throw means those were rolled
+    /// back.
+    func setAudioBitrateProfile(_ profile: AudioBitrateProfile) async throws {
+        guard let call, await call.state.session != nil else {
+            throw ClientError("Audio bitrate profiles require a joined call.")
+        }
+        try await ensureHiFiAudioEnabled()
```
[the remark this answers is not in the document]
ipavlidakis: `.voiceStandard` now skips the dashboard hi-fi check so a mid-call `hifi_audio_enabled = false` can still restore voice. Music and voice-high-quality still require the entitlement. Covered in `test_setAudioBitrateProfile_hifiDisabled_*`. Fixed in 4813ca4d.

Sources/StreamVideo/WebRTC/v2/PeerConnection/MediaAdapters/LocalMediaAdapters/LocalAudioMediaAdapter.swift
```diff
@@ -336,6 +338,21 @@ final class LocalAudioMediaAdapter: LocalMediaAdapting, @unchecked Sendable {
         with layerSettings: [Stream_Video_Sfu_Event_AudioSender]
     ) { /* No-op */ }
 
+    func setMaxBitrate(for profile: AudioBitrateProfile) async {
+        try? await processingQueue.addSynchronousTaskOperation { [weak self] in
+            guard let self else { return }
+            audioBitrateProfile = profile
+            transceiverStorage.forEach { options, value in
+                if let bitrate = bitrate(for: options, profile: profile) {
+                    applyMaxBitrate(bitrate, on: value.transceiver)
```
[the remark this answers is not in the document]
ipavlidakis: Bitrate now resolves against the live `publishOptions` entry (id+codec match), and a publish-options refresh re-stamps the active profile. Added `test_setMaxBitrate_publishOptionsProfileMapChange_usesCurrentBitrate`. Fixed in 4813ca4d.

Sources/StreamVideo/WebRTC/v2/WebRTCStateAdapter.swift
```diff
@@ -485,6 +527,12 @@ actor WebRTCStateAdapter: ObservableObject, StreamAudioSessionAdapterDelegate, W
         try await restoreScreenSharing()
         publisher.setVideoFilter(videoFilter)
         publisher.completeSetUp()
+        // Rebind: new senders need the live bitrate even if music is
+        // already on. Session is re-asserted; VP no-ops if unchanged.
+        try await applyAudioBitrateProfile(
+            audioBitrateProfile,
+            context: .rebind
+        )
```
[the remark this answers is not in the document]
ipavlidakis: Rebind failures are logged and no longer abort `configurePeerConnections` before `subscriber.setUp`. The applicator still rolls the profile back. Fixed in 4813ca4d.

StreamVideoTests/Call/Call_Tests.swift
```diff
@@ -1188,6 +1188,48 @@ final class Call_Tests: StreamVideoTestCase, @unchecked Sendable {
...
+        let callController = MockCallController()
+        let call = Call.dummy(callController: callController)
+
+        call.update(reconnectionStatus: .connected)
+        await fulfilmentInMainActor {
+            call.state.reconnectionStatus == .connected
+        }
+        call.state.session = .dummy()
+        call.setAudioFilter(noiseCancellationFilter)
+        streamVideo.videoConfig.audioProcessingModule.setAudioFilter(nil)
```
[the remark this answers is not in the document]
ipavlidakis: The shared processing-module reset now runs in `defer`, so a failed assertion or fulfilment timeout cannot leak the filter into later tests. Fixed in 4813ca4d.

StreamVideoTests/CallSettings/MicrophoneManager_Tests.swift
```diff
@@ -96,4 +96,113 @@ final class MicrophoneManager_Tests: XCTestCase, @unchecked Sendable {
         XCTAssertEqual(subject.status, .disabled)
         XCTAssertEqual(callController.timesCalled(.changeAudioState), 0)
     }
+
+    // MARK: - setAudioBitrateProfile
+
+    func test_setAudioBitrateProfile_updatesPublishedValueOnSuccess() async throws {
```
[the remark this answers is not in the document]
ipavlidakis: Leaving these names as-is. The new methods already match the existing `test_<setup>_<action>_<result>` pattern in these files. The trailing `_()` suffix is not applied in StreamVideoTests today, including pre-existing methods in the same classes.

StreamVideoTests/CallSettings/AudioBitrateProfile_Tests.swift
```diff
@@ -0,0 +1,50 @@
+//
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamVideo
+import XCTest
+
+final class AudioBitrateProfile_Tests: XCTestCase, @unchecked Sendable {
+
+    func test_defaultBitrate_matchesVoiceAndMusicRates() {
```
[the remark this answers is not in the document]
ipavlidakis: Leaving these names as-is. They already follow `test_<setup>_<action>_<result>`. The trailing `_()` suffix is not used by the existing tests in this file or the rest of StreamVideoTests, and renaming would be churn without changing coverage.

DemoApp/Sources/ViewModifiers/MoreControls/DemoMusicModeButtonView.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    @ObservedObject var viewModel: CallViewModel
+
+    var body: some View {
+        if let call = viewModel.call {
+            Content(microphone: call.microphone)
+        }
+    }
+
+    private struct Content: View {
```
martinmitrevski, line 20: Seems a bit unnecessary? Why not put the button directly in the body?

Sources/StreamVideo/CallSettings/MicrophoneManager.swift
```diff
@@ -59,6 +62,38 @@ public final class MicrophoneManager: ObservableObject, CallSettingsManager, @un
...
+    /// cached `Call` is not a no-op. Reconnect re-applies bitrate on the
+    /// new publisher.
+    ///
+    /// - Parameter profile: Voice or music capture profile.
+    /// - Throws: `ClientError` when the call is missing, hi-fi is off on
+    ///   the dashboard, or Voice Processing cannot be applied.
+    public func setAudioBitrateProfile(_ profile: AudioBitrateProfile) async throws {
+        try await audioBitrateProfileQueue.addSynchronousTaskOperation { [self] in
+            let current = await MainActor.run { audioBitrateProfile }
+            guard profile != current else { return }
```
martinmitrevski: This guard is the only place a second copy of the profile can go stale, and it can strand the API.

resetAudioBitrateProfile() is called only from Call+LeavingStage (line 110), which is only reached via Call.leave(). But WebRTCStateAdapter.cleanUp() — which resets the applicator to .voiceStandard — is also reached from WebRTCCoordinator+Error and WebRTCCoordinator+Blocked, and Call.StateMachine.Stage.ErrorStage just logs and transitions to .idle.

So after an unrecoverable reconnect or a block while music is on: hardware is correctly back on voice, but audioBitrateProfile still publishes .musicHighQuality. The published value is wrong, and the next setAudioBitrateProfile(.musicHighQuality) returns here without applying anything.

AudioBitrateProfileApplicator.applySerially already no-ops on context == .setProfile, profile == previous, so this guard is redundant. Could we drop it and set the published value from the applicator's profile after the apply returns? That removes the divergence class rather than patching the two extra teardown paths, which is the version that breaks again when a third one is added.

Sources/StreamVideo/WebRTC/v2/VideoCapturing/ActionHandlers/ScreenShare/ScreenShareCaptureHandler.swift
```diff
@@ -249,21 +264,22 @@ final class ScreenShareCaptureHandler: NSObject, StreamVideoCapturerActionHandle
...
         else {
             return
         }
 
-        // Restore the previously disabled filter
-        audioFilterProcessingModule.setAudioFilter(audioFilterBeforeScreensharingAudio)
-        audioFilterBeforeScreensharingAudio = nil
-
+        defer {
+            if !recorder.isRecording {
```
martinmitrevski: Gating the release on !recorder.isRecording means that if stopCapture() on line 282 throws while the recorder is still recording, the gate is never released. applyLiveFilter() then keeps writing nil to the processing module, so the integrator's audio filter — including noise cancellation — stays off for the rest of the call, and isRecording stays true so nothing here re-runs.

The pre-PR code restored the filter unconditionally before stopCapture(), so this is a regression in the error path. Can we call audioFilterGate?(false) unconditionally in the defer and keep only activeSession/isRecording behind the recorder.isRecording check?

Sources/StreamVideo/Utils/AudioSession/AudioDeviceModule/AudioDeviceModule.swift
```diff
@@ -259,22 +261,161 @@ final class AudioDeviceModule: NSObject, RTCAudioDeviceModuleDelegate, Encodable
...
+                ),
+                restorePlayout: restorePlayout
+            )
+            musicCapturePolicy.markApplied(
+                stereoPreferred: source.prefersStereoPlayout
+            )
+        } catch {
+            try? applyCapturePolicyLocked(
+                musicCapturePolicy.appliedConfiguration,
+                restorePlayout: false
```
martinmitrevski: applyDeferredCapturePolicyLocked keeps pendingPlayoutRestore until the apply succeeds, but the direct unmuted path doesn't. If stopPlayout() succeeds and startPlayout() then fails, this rollback runs with restorePlayout: false and the catch at line 328 restores previousPolicy, dropping the obligation entirely.

Playout is now down, so isPlaying (driven by the engine delegate callbacks) and source.isPlaying are both false. An identical retry computes shouldRestorePlayout == false, applies cleanly, and reports success with remote audio still stopped. Even without the retry, the first failure leaves playout dead and nothing here brings it back.

Can we carry the obligation until startPlayout() succeeds? Note that pendingPlayoutRestore can't be reused as-is: hasPendingApply is derived from it, so setting it after a rollback would make the next unmute re-apply the rolled-back policy. It likely needs splitting into "an apply is deferred" and "playout still owes a restart." Restoring playout inside the rollback is the other option, but only when the throw actually came from resetPlayoutLocked — for a setVoiceProcessingEnabled failure, playout was never stopped and false is correct.

Sources/StreamVideo/CallSettings/MicrophoneManager.swift
martinmitrevski, line 82: This guard makes MicrophoneManager’s independently cached published value an execution gate. That cache can diverge on the blocked-user path: CallController transitions the coordinator directly to .blocked, coordinator cleanup resets the applicator to .voiceStandard, but .blocked is not bridged to Call.leave(), so audioBitrateProfile still publishes music. A later setAudioBitrateProfile(.musicHighQuality) then returns here without applying it.
The normal .error path is covered because CallController explicitly calls call.leave() for .error. Could we make the published value follow the applicator/controller’s committed profile, or explicitly synchronize it during blocked cleanup? Removing this guard is useful defense, but by itself it does not repair the stale published value.

Sources/StreamVideo/WebRTC/v2/VideoCapturing/ActionHandlers/ScreenShare/ScreenShareCaptureHandler.swift
```diff
@@ -249,21 +264,22 @@ final class ScreenShareCaptureHandler: NSObject, StreamVideoCapturerActionHandle
...
+                audioFilterGate?(false)
```
martinmitrevski, line 277: RPScreenRecorder.stopCapture invokes its completion when capture has stopped, including when it supplies an error. This async wrapper resumes only from that completion, but the defer then treats recorder.isRecording as a second source of truth. If that property is lagging and still reads true, the gate, activeSession, and the handler’s isRecording flag are all left active, so live audio filters remain suppressed.
Could we centralize terminal cleanup in a finishCapture() helper and invoke it unconditionally once the stop completion returns or throws? The didStopRecordingWith delegate callback should use the same finalizer—even when error == nil—rather than trying to stop an already-stopped recorder again. A stop-error test with recorder.isRecording == true would cover the regression.

Sources/StreamVideo/Utils/AudioSession/AudioDeviceModule/AudioDeviceModule.swift
martinmitrevski, line 398: If stopPlayout() succeeds and the following start fails, this rollback restores the capture configuration with restorePlayout: false, and the outer catch at line 328 then restores previousPolicy. That loses the fact that playout still owes a restart. Once the engine/native state reports stopped, an identical retry computes shouldRestorePlayout == false and can return success while remote audio remains down.
Could we track restart debt separately from “capture apply is deferred” and clear it only after playout starts successfully, or perform a compensating restart when the failure occurred after a successful stop? pendingPlayoutRestore cannot safely represent both states because it also drives hasPendingApply.

Sources/StreamVideo/WebRTC/v2/PeerConnection/MediaAdapters/LocalMediaAdapters/LocalAudioMediaAdapter.swift
```diff
@@ -366,6 +380,67 @@ final class LocalAudioMediaAdapter: LocalMediaAdapting, @unchecked Sendable {
...
+        on transceiver: RTCRtpTransceiver
+    ) {
+        let params = transceiver.sender.parameters
+        if bitrate <= 0 {
+            guard !params.encodings.isEmpty else { return }
+            params.encodings.forEach { $0.maxBitrateBps = nil }
+            transceiver.sender.parameters = params
+            return
+        }
+        if params.encodings.isEmpty {
```
martinmitrevski, line 435: The sender-parameter contract requires the number of encodings to remain equal to the sender’s existing send-encoding count; changing it from 0 to 1 is rejected as an invalid modification. Libwebrtc also initializes senders with a default encoding.
That makes this branch either unreachable on a normal sender or ineffective when the binding actually exposes an empty list, and this property assignment gives us no error to act on. Could we remove the synthesis and rely on the post-transceiver apply, or add an integration test using a real RTCRtpSender that assigns the parameters and then reads maxBitrateBps back? The comment above should not promise that the cap is preserved without that verification.

martinmitrevski (commented):
Looks good, left few comments

DemoApp/Sources/ViewModifiers/MoreControls/DemoMusicModeButtonView.swift
ipavlidakis: Leaving the nested `Content` view. It exists so we can `@ObservedObject` the `MicrophoneManager`. Putting the button in `body` would not refresh when `audioBitrateProfile` changes unless `CallViewModel` republished that value.

Sources/StreamVideo/CallSettings/MicrophoneManager.swift
ipavlidakis: Valid. Blocked cleanup never goes through `Call.leave()`, so the published value stayed at music while the applicator reset to voice.

The same-profile guard stays (it is the right no-op once the cache is correct). Blocked-user handling now calls `resetAudioBitrateProfile()` next to the coordinator `.blocked` transition, same serial queue as leave.

Sources/StreamVideo/WebRTC/v2/VideoCapturing/ActionHandlers/ScreenShare/ScreenShareCaptureHandler.swift
ipavlidakis: Valid. `stopCapture` completion is the source of truth, including when it supplies an error. Cleanup is now an idempotent `finishCapture()` from both `stop()` (via `defer`) and `didStopRecordingWith` (even when `error == nil`), without calling `stopCapture` again or consulting `recorder.isRecording`.

Sources/StreamVideo/WebRTC/v2/PeerConnection/MediaAdapters/LocalMediaAdapters/LocalAudioMediaAdapter.swift
ipavlidakis: Valid. The empty-list synthesis is gone. We only write `maxBitrateBps` on existing encodings; an empty list is left unchanged because changing the send-encoding count from 0 to 1 is an invalid modification. New transceivers still get the cap from `addTransceiverIfRequired`.

Sources/StreamVideo/Utils/AudioSession/AudioDeviceModule/AudioDeviceModule.swift
ipavlidakis: Valid. Restart debt is now `playoutRestartOwed` on the ADM, not on `MusicCapturePolicy`. It is set after a successful `stopPlayout` and cleared only after start succeeds. Capture-policy rollback still restores VP with `restorePlayout: false`, then tries a compensating start if debt is owed. A retry with the engine already stopped still restarts playout.

martinmitrevski (approved):
LGTM ✅

PR #1271, opened 2026-09-11
Tokenize the Video lobby

nuno-vieira:
### 🔗 Issue Links

- [IOS-2015](https://linear.app/stream/issue/IOS-2015/tokenize-the-video-lobby-and-stabilize-call-control-icon-transitions)

**Stack:** this PR (`v2-lobby-design-refresh` → `v2-core-token-namespaces`) is the base. Follow-up [#1272](https://github.com/GetStream/stream-video-swift/pull/1272) (`v2-call-icon-font-sizing` → this branch) should merge after this one.

### 🎯 Goal

Token-migrate the Video lobby to `VideoAppearance` without matching Figma layout. Figma is only a lookup for which tokens the existing surfaces should use.

### 📝 Summary

- Drive lobby spacing, sizes, and colors from `videoAppearance.tokens` / `videoAppearance.colors`
- Style mic and camera controls for on, off, and permission-denied using Figma-assigned tokens
- Expose Video icons on `videoAppearance.images`, including the permission-error badge
- Document the v2 design-refresh approach in `AGENTS.md`

### 🛠 Implementation

Lobby layout is unchanged. Shared controls (`ToggleControlStyle.permissionDenied`, `ControlBadgeView`, `StatelessMicrophoneIconView` / `StatelessVideoIconView`) now read tokens so off and permission-denied states match Figma semantics. Icons come from `videoAppearance.images` instead of inlined SF Symbols.

This PR does **not** change `CallIconView` sizing. That lives in stacked [#1272](https://github.com/GetStream/stream-video-swift/pull/1272).

### 🎨 Showcase

Git diffs of existing lobby / control snapshots plus new lobby snapshots for mic+camera off and permission denied.

|  Before  |  After  |
| -------- | ------- |
|  existing lobby snapshots  |  tokenized lobby snapshots in this PR  |

### 🧪 Manual Testing Notes

- Lobby with mic and camera on, permissions granted
- Lobby with mic and camera off
- Lobby with mic and camera permission denied (error badge, disabled style)
- Confirm existing lobby layout (headers, participant cards, join CTA) is unchanged aside from token colors/spacing
- Run `StreamVideoSwiftUI` snapshot suites for `LobbyView` and call controls

### ☑️ Contributor Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
[...]

PR #1272, opened 2026-09-11
Stabilize call control icon transitions

nuno-vieira:
### 🔗 Issue Links

- [IOS-2015](https://linear.app/stream/issue/IOS-2015/tokenize-the-video-lobby-and-stabilize-call-control-icon-transitions)

**Stack:** this PR (`v2-call-icon-font-sizing` → `v2-lobby-design-refresh`) sits on [#1271](https://github.com/GetStream/stream-video-swift/pull/1271). Merge #1271 first; this PR only contains the icon-transition fix.

### 🎯 Goal

Stop call-control icons from changing apparent size when toggling on/off. Slash SF Symbols have a taller bounding box, and scale-to-fit was shrinking/growing the glyph during the animation.

### 📝 Summary

- Size call-control SF Symbols with a tokenized font size (`tokens.layout.iconSizeSm`) instead of `.resizable()` + a max frame
- Prevents slash variants (`video.slash.fill`, `mic.slash.fill`) from jumping during toggle animations
- Custom non-symbol images keep their intrinsic size; `.font` only applies to SF Symbols

### 🛠 Implementation

`CallIconView` is shared by lobby and in-call controls. Font sizing draws both symbol variants at a fixed point size so the camera/mic body stays put and only the slash is added. Width/height scale-to-fit and stretching were tried first; they either kept the jump or distorted the glyphs.

### 🎨 Showcase

Git diffs of `CallIconView` consumer snapshots (call controls, lobby, connecting/outgoing/joining).

|  Before  |  After  |
| -------- | ------- |
|  existing control snapshots  |  font-sized control snapshots in this PR  |

### 🧪 Manual Testing Notes

- Toggle camera on/off in lobby and in-call controls — glyph should not grow/shrink
- Toggle microphone on/off — same
- Hang up, speaker, and participants icons still look correctly sized
- Snapshot suites that render `CallIconView`

### ☑️ Contributor Checklist

### 🎁 Meme

_n/a_


Sources/StreamVideoSwiftUI/Utils/HelperViews.swift
```diff
@@ -33,9 +38,7 @@ public struct CallIconView: View {
                 iconStyle.backgroundColor.opacity(iconStyle.opacity)
             )
             icon
-                .resizable()
-                .aspectRatio(contentMode: .fit)
-                .frame(maxWidth: 22, maxHeight: 20)
```
martinmitrevski, line 38: hmm, could removing the frame be a problem for non-sf symbols?

nuno-vieira: This icon is universal, I would doubt that any customer would want to change the microphone or camera icons. But if we would like to support this properly, then it would most likely make sense to export it as an actual view. In a way, having the `Appearance.Images` objects is a bit limited, because customers can't control much, the can select an image,  but we won't allow customers select between fonts or resizable images. Only an actual view would be that flexible

nuno-vieira: If customers used a regular image, it would still be mostly fine because of the outer frame size. At most, they could use an image with the UIKit init `Image(uiImage:)` and provide a size there. Or later we can also provide a view slot for this if needed.

ipavlidakis: I'm not sure about this one. The icons should remain generic. Resizable, frame etc are things that require context. Our icons shouldn't carry that information. 

For example, the same icon can be used in a button but it can also be used in a list a list of calls to that that this was an answered call (where in the list it can be smaller).

My point is, the icons themselves shouldn't carry the appearance configuration which belongs in the component.

ipavlidakis:
I'm not up for this change. I think it tries to provide solution to an unproven problem while making things less clear and moving rendering responsibilities into the Image token themselves.

PR #1276, opened 2026-09-15
[Enhancement]Reduce SDK size

martinmitrevski:
### 🔗 Issue Links

Addresses the "Reduce SDK size" item on the README roadmap.

### 🎯 Goal

Cut the shipped size of the SDK without changing behaviour or breaking any public API.

### 📝 Summary

Measured by archiving `DemoAppUIKit` in Release and reading the embedded framework binaries — the same basis the SDK Size job uses — for `develop` (`2c008182`) and this branch on the same machine and toolchain:

| | develop | branch | diff | |
|---|---:|---:|---:|---:|
| StreamVideo | 9,584,440 | 9,535,456 | **−48,984** | −0.51% |
| StreamVideoSwiftUI | 2,465,136 | 2,465,136 | 0 | 0.00% |
| StreamVideoUIKit | 106,840 | 106,840 | 0 | 0.00% |
| ringtone resource | 717,180 | 14,330 | **−717,180** | −98% |
| **total app payload** | | | **−766,164** | |

**Note the SDK Size job will under-report this PR.** `frameworks_sizes` measures `File.size(".../StreamVideoSwiftUI.framework/StreamVideoSwiftUI")` — the Mach-O binary, not the bundle — so the 700KB resource saving, which is the bulk of the win, is invisible to it. Expect the bot to show roughly −48KB on StreamVideo and ~0 elsewhere, while an app embedding the SDK actually shrinks by ~748KB.

### 🛠 Implementation

- The incoming ringtone was 714KB of uncompressed 48kHz stereo PCM. Re-encoded to AAC, same codec and container as the existing `outgoing.m4a`: **731,510 → 14,330 bytes**, same 3.81s duration, 48kHz stereo, 30.8dB SNR on round-trip. `Sounds.incomingCallSound` now defaults to `"incoming.m4a"`.
- `SelectiveEncodable.encode(to:)` had its reflection loop cloned for all 38 conforming protobuf types (54,108 bytes of specialisations). The body moved into a non-generic `@inline(never)` function; the conformances are now thin thunks. The SDK Size job's XCSize breakdown confirmed this at `Protobuf+SelectiveEncodable.o −52,757`.
- `OneOf_EventPayload.payload(_:)` was a 23-case switch cloned 22 times (12,568 bytes). Extracted into a non-generic `boxedPayload: Any`.

#### Where the size actually goes

Link-map attribution of `StreamVideo` on `develop`: statically linked SwiftProtobuf 1.17MB (13.8%), generated `.pb.swift` 563KB, generated `DefaultAPI` 180KB, `Codable` synthesis ~883KB, `Hashable`/`Equatable` synthesis ~502KB. Most of that is public API surface or generated code that cannot move without a breaking change.

#### Considered and rejected

- **`-Osize` for the SDK targets in Release.** Worth −392KB on StreamVideo and −49KB on StreamVideoSwiftUI, but it reaches no SPM consumer — apps that integrate via SPM compile the package with their own build settings, so it would move the size badge and the demo apps and nothing else. Dropped rather than take an inlining trade-off on a real-time media SDK for a metrics-only gain. Easy to revisit: it is one line per target.
- **Hiding SwiftProtobuf's 5,130 exported symbols** so `-dead_strip` could reach the unused TextFormat / JSON / descriptor code. Measured at only **−144KB**, because Swift conformance records in `__swift5_proto` are non-strippable dead-strip roots. Not worth a glob-based linker hack.
- **Removing dead internal code** (`RetriableTask`, `TrailingView`, `TopView`). Genuinely unused, but unused generic code emits almost nothing, so it is churn without benefit.

### 🧪 Manual Testing Notes

- `SelectiveEncodable` JSON output compared before and after across seven representative SFU messages — **byte-identical** — plus payload-cast parity checks.
- Confirmed `incoming.m4a` ships in the built `StreamVideoSwiftUI.framework` and loads through `AVAudioPlayer` at the same duration.
- `WebRTCTrace_Tests`, `WebRTCTracesAdapter_Tests`, `SFUEventAdapter_Tests`, `SFUAdapter_Tests`, `SFUAdapterEvent_Tests` all pass.
- `swiftformat --lint` and `swiftlint` clean.

Two pre-existing failures reproduce identically on a clean `develop` on my machine and are unrelated to this PR: `WebRTCCoordinatorStateMachine_JoiningStageTests.test_willTransitionAway_audioReadinessPending_skipsPeerConnections`, and 49 `StreamVideoSwiftUITests` snapshot tests whose references were recorded on different simulator hardware. I verified every one of the 49 against a stashed tree — the failure sets match test for test.

### ⚠️ One call for the reviewer

The ringtone change is a default-value change, not an API break. Apps that set `Sounds.incomingCallSound` themselves are unaffected. One edge case: an app that overrides only `Sounds.bundle` and ships its own `incoming.wav` would now get no incoming sound. Called out in the CHANGELOG.

Unrelated find, not addressed here: `L10n_template.stencil` (a SwiftGen build-time input) is being copied into the shipped `StreamVideoSwiftUI.framework`. Only 3KB, but it does not belong in the product.

### ☑️ Contributor Checklist


martinmitrevski:
handled in a different PR

PR #1279, opened 2026-09-16
[Fix] ADM crash on leave, REST-only LLC, and local pins

ipavlidakis:
### 🔗 Issue Links

N/A

### 🎯 Goal

Stop a native AudioDeviceModule crash when a call leaves or another call takes over the process-wide audio store. Keep LLC tests from burning the timeout budget on coordinator WebSockets they do not need. Keep local pins when the SFU sends PinsChanged (JS parity). Make CleanUpStage close recording reliable under the full suite.

### 📝 Summary

**ADM teardown on leave**

- Do not `reset()` the previous ADM on store swap or nil. That rebuilds the native AudioEngine graph while WebRTC workers can still be inside it.
- `cleanUp` closes publisher/subscriber first, then `deactivate()`. Deactivate stays session-leased, so a replacement call that already owns the store is not torn down.
- Do not call `setEngineAvailability(false)` in `cleanUp`. Availability is process-wide and would mute a call that already claimed the engine.
- Drain leftover local-audio publish (`stop()` / `isStopped`) before native peer-connection close, so mute-RPC → clone/`makeAudioTrack` cannot run after Call deinit.
- `StreamRTCPeerConnection.close()` awaits MainActor transceiver-stop + close instead of fire-and-forget.
- Nil the native ADM observer in `AudioDeviceModule.deinit`.
- Hold publisher/subscriber across the profile-reset await so that hop cannot skip `close()` on the connections this session still owns.

**Local pins vs PinsChanged**

- `handlePinsChanged` matches JS `CallState.setServerSidePins`: keep local pins, apply a server pin only when the participant has none, clear only remote pins that left the SFU list.

**REST-only LLC / timeout budget**

- `callFlow` accepts `connectMode`. Device/list/edges tests use `.none` so they skip coordinator WS. The two `get()` error tests keep connecting so `APIError` wrapping matches production.
- Cap client disconnect during teardown at 2s. Disable mic/camera before `leave()`. Wait for the store ADM to become nil instead of dispatching nil.
- Restore process-wide `STREAM_TESTS` timeouts after LLC teardown.
- Serialize ringing `create` before `accept`; create audio-room participant `callFlow`s before the task group.

**Join capabilities / subscriptions**

- Skip the adapter’s empty pre-join `ownCapabilities` seed so it cannot overwrite `JoinCallResponse` capabilities. Late subscriptions still apply a real set (no `dropFirst()`).
- Forward joined-stage capability updates on unique task identifiers so a later emission does not cancel the join-time value.
- Subscribe to track updates synchronously so the first `updateSubscriptions` is not delayed onto `processingQueue`.

**Tests / mocks**

- Record `MockRTCPeerConnectionCoordinator` calls via `Atomic.mutate` so concurrent appends cannot drop `close`.
- CleanUpStage waits for idle instead of a 0.5s sleep, then asserts `close` was called.
- Box screenshare/zoom mock payloads (`BeginScreenSharingPayload`, `ZoomPayload`) so `recordedInputPayload` round-trips through `Any`.

### 🛠 Implementation

The native ADM is process-wide. `reset()` and unconditioned `setEngineAvailability(false)` on leave are unsafe when another call may already own the store, and they race worker threads still in the graph. Teardown is now: drain leftover publish → await PC close → drop the store pointer if this session still owns it. The engine graph dies with that close, not with `reset()`.

Coordinator WS connect is ~30s × retries and was the LLC Debug timeout. HTTP still works without that socket.

JS keeps local pins when the SFU pin list changes. Swift used to wipe any pin not on that list, which hung/failed the local-pin LLC tests. `handlePinsChanged` now follows the JS policy.

`@Atomic` get-then-append is two operations; under full-suite load that dropped `timesCalled(.close)`. Recording goes through `mutate`. CleanUpStage waits for the idle transition instead of sleeping.

The StreamVideo scheme also turns off Metal API validation (Xcode launch setting). That is unrelated to ADM/LLC/pin behavior.

### 🎨 Showcase

N/A

### 🧪 Manual Testing Notes

Local Xcode StreamVideo test plan (Test configuration). Full suite on `cb6b8c03`: 2181 passed, 6 skipped, 0 failed.

### ☑️ Contributor Checklist

- [x] I have signed the Stream CLA (required)


Sources/StreamVideo/WebRTC/v2/PeerConnection/MediaAdapters/MediaAdapter.swift
```diff
@@ -349,9 +349,14 @@ final class MediaAdapter {
     ) async throws {
         try await videoMediaAdapter.removeCapturePhotoOutput(capturePhotoOutput)
     }
-    
+
+    /// Waits until local audio publish has finished or been cancelled.
+    func stopLocalAudio() async {
```
ipavlidakis: rather than stopLocalAudio we can rename it stopMedia and make it a method for all localAdapters while screenshare and camera can simply ignore it

StreamVideo.xcodeproj/xcshareddata/xcschemes/StreamVideo.xcscheme
```diff
@@ -75,6 +75,9 @@
       debugServiceExtension = "internal"
       enableGPUValidationMode = "1"
       allowLocationSimulation = "YES">
+      <MetalAPIValidationSettings
```
ipavlidakis, line 78: This is being added by Xcode 27 and there doesn't seem to be a way to disable it

StreamVideoTests/IntegrationTests/Components/Helpers/Call_IntegrationTests+Helpers.swift
```diff
@@ -104,18 +107,25 @@ extension Call_IntegrationTests {
...
 
-        // Production 10s. No-op if never stubbed.
+        // Helpers.init applies production timeouts for live API calls.
+        // Put the process-wide STREAM_TESTS defaults back so later
+        // tests on this xctest clone do not inherit production values.
         func restoreAudioSessionReadinessWatchdog() {
-            WebRTCConfiguration.timeout.audioSessionReadinessWatchdog =
-                WebRTCConfiguration.Timeout.production
-                    .audioSessionReadinessWatchdog
+            #if STREAM_TESTS
```
ipavlidakis, line 114: That can be simplified as tests always run with STREAM_TESTS defined and equal to 1.  Just use directly assignments to .testing

